PR #15, opened 2024-12-11
Set up form validation

chelseaSchmidt:
The bones of form validation! Including:

- Introducing an overall validation pattern
- Simple validation of top-level required fields, to demonstrate the pattern works

It's very long, sorry. Once this foundation is in it should be easier to make PRs more bite-sized 😬

### DIFF
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -146,12 +146,12 @@ const initialFormData: FormData = {
     competitionTypes: {
         value: [],
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     league: {
         value: null,
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     usedExpansions: {
         value: null,
@@ -163,12 +163,12 @@ const initialFormData: FormData = {
     expansions: {
         value: [],
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     wasTreebeardMustered: {
         value: null,
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     usedHandicap: {
         value: null,
@@ -177,8 +177,8 @@ const initialFormData: FormData = {
             return detectMissingInput(this.value);
         },
     },
-    actionTokens: { value: 0, error: null, validate: () => null },
-    dwarvenRings: { value: 0, error: null, validate: () => null },
+    actionTokens: { value: 0, error: null, validate: alwaysValid },
+    dwarvenRings: { value: 0, error: null, validate: alwaysValid },
     gameTurns: {
         value: UNINITIALIZED,
         error: null,
@@ -203,7 +203,7 @@ const initialFormData: FormData = {
     mordorTrack: {
         value: UNINITIALIZED,
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     initialEyes: {
         value: UNINITIALIZED,
@@ -222,12 +222,12 @@ const initialFormData: FormData = {
     aragornCrownedTurn: {
         value: UNINITIALIZED,
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     capturedStrongholds: {
         value: [],
         error: null,
-        validate: () => null,
+        validate: alwaysValid,
     },
     interestRating: {
         value: -1,
@@ -236,7 +236,7 @@ const initialFormData: FormData = {
             return detectMissingInput(this.value);
         },
     },
-    comment: { value: "", error: null, validate: () => null },
+    comment: { value: "", error: null, validate: alwaysValid },
 };
 
 type ValueOf<T> = T[keyof T];
@@ -926,6 +926,10 @@ function detectMissingInput(value: unknown): FieldError {
         (typeof value === "number" ? value >= 0 : true)
         ? null
         : ErrorMessage.Required;
+}
+
+function alwaysValid() {
+    return null;
 }
 
 function objectKeys<T extends object>(obj: T): Array<keyof T> {
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -257,10 +257,12 @@ function GameReportForm() {
         return () => {
             setFormData((prevData) => {
                 const error = prevData[field].validate();
-                return {
-                    ...prevData,
-                    [field]: { ...prevData[field], error },
-                };
+                return error
+                    ? {
+                          ...prevData,
+                          [field]: { ...prevData[field], error },
+                      }
+                    : prevData;
             });
         };
     };
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -110,133 +110,133 @@ const INFINITE = 100;
 const initialFormData: FormData = {
     winner: {
         value: null,
-        error: "",
+        error: null,
         validate: function _() {
             return detectMissingInput(this.value);
         },
     },
     loser: {
         value: null,
-        error: "",
+        error: null,
         validate: function _() {
             return detectMissingInput(this.value);
         },
     },
     side: {
         value: null,
-        error: "",
+        error: null,
         validate: function _() {
             return detectMissingInput(this.value);
         },
     },
     victoryType: {
         value: null,
-        error: "",
+        error: null,
         validate: function _() {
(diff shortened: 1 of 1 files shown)


frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
-    corruption: UNINITIALIZED,
-    didFellowshipReachMordor: null,
-    mordorTrack: UNINITIALIZED,
-    initialEyes: UNINITIALIZED,
-    wasAragornCrowned: null,
-    aragornCrownedTurn: UNINITIALIZED,
-    capturedStrongholds: [],
-    interestRating: -1,
-    comment: "",
+    winner: {
```
chelseaSchmidt, line 111: The change to `FormData` even on its own adds a silly number of lines - in the next PR I want to split things off `GameReportForm` into smaller files, but I figured if I did it now it would be harder to review the diff.

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
+                        [field]: { ...prevData[field], error: fieldError },
+                    }));
+                }
+            });
+            return formError;
+        }
+        return null;
+    };
+
+    const toPayload = (formData: FormData) =>
```
chelseaSchmidt, line 287: This only does the one layer of cleaning, which is stripping out the frontend-only data; saving redundant field cleanup for a separate PR (it should be small I think)

frontend/src/GameReportForm.tsx
```diff
@@ -218,224 +397,317 @@ function GameReportForm() {
...
+function GameReportFormElement({
+    children,
+    label,
+    error,
+}: GameReportElementProps) {
     return (
         <Sheet variant="outlined" sx={{ p: 2, borderRadius: "lg" }}>
-            <FormLabel sx={{ fontSize: 16, pb: 2 }}>{label}</FormLabel>
-            {children}
+            <FormControl error={!!error}>
```
chelseaSchmidt, line 706: I added a `FormControl` wrapper to take advantage of MUI's default error styling, but it does introduce a console error `A FormControl can contain only one control component...` because of how one of the multiselect list components is implemented. Want to fix separately so I don't mix in a lot of UI changes

frontend/src/GameReportForm.tsx
```diff
@@ -625,4 +911,28 @@ function VictoryPoints({ strongholds }: VictoryPointsProps) {
     );
 }
 
+function detectMissingInput(value: unknown): FieldError {
+    return value !== null &&
+        value !== undefined &&
+        value !== "" &&
+        (typeof value === "number" ? value >= 0 : true)
```
chelseaSchmidt, line 918: Not in this PR, but maybe `number` fields should be changed to `number | null` instead of having `-1` represent an empty input? That would be parallel to the `string | null` fields. Alternatively, could make the `string | null` fields only `string` and have `""` represent an empty input for strings. Any preference?

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
+                          ...prevData,
+                          [field]: { ...prevData[field], error },
+                      }
+                    : prevData;
+            });
+        };
     };
 
+    const validateForm = () => {
+        const formError = detectFormErrors(formData);
```
sirrus233, line 271: I'm not sure what the point is of pre-detecting errors in `detectFormErrors` if we are then going to re-call validate on every field inside the loop? Maybe I'm just missing it.

frontend/src/GameReportForm.tsx
sirrus233: Yes, 100%. I'm not sure why I did it this way, probably because I was trying to keep some fields non-nullable. But I was conflating UI representation (e.g. `gameTurns` could be `null`) with data representation (`gameTurns` is a non-null int >= 1). That can be fixed now.

I probably prefer null for no input (absence of a value)? Just a slight preference because it mirrors the other fields that are missing their values. No strong opinion though.

sirrus233 (approved):
Glad you're tackling this and ripping it apart as appropriate!

Left some comments, but approving so I don't get in your way (because I know there's a lot to refactor here). Feel free to merge when you think it's ready!

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
-    mordorTrack: UNINITIALIZED,
-    initialEyes: UNINITIALIZED,
-    wasAragornCrowned: null,
-    aragornCrownedTurn: UNINITIALIZED,
-    capturedStrongholds: [],
-    interestRating: -1,
-    comment: "",
+    winner: {
+        value: null,
+        error: "",
```
sirrus233, line 113: Shouldn't the initialized error fields be `null`?

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
+        validate: function _() {
+            return detectMissingInput(this.value);
```
sirrus233, line 115: After researching, I guess this idiom is okay. This is just a terrible programming language :wink:

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
+        },
+    },
+    competitionTypes: {
+        value: [],
+        error: "",
+        validate: () => null,
```
sirrus233, line 149: Want to name this, for clarity? Perhaps `const alwaysValid = () => null` ?

frontend/src/GameReportForm.tsx
chelseaSchmidt: Omg this made me laugh 🤣

frontend/src/GameReportForm.tsx
chelseaSchmidt: Like the idea of conceptual overhead 💯

frontend/src/GameReportForm.tsx
chelseaSchmidt: Been there haha. With you on `null`, let's do it!

frontend/src/GameReportForm.tsx
```diff
@@ -65,93 +65,263 @@ const cities: Stronghold[] = [
...
+        },
+    },
+    comment: { value: "", error: "", validate: () => null },
 };
 
 type ValueOf<T> = T[keyof T];
 
 function GameReportForm() {
     const [formData, setFormData] = useState(initialFormData);
+    const [errorOnSubmit, setErrorOnSubmit] = useState<string | null>(null);
```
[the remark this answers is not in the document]
chelseaSchmidt: Yup 🤦‍♀️

frontend/src/GameReportForm.tsx
chelseaSchmidt: My bad, it's unnecessary, you're not missing anything! It was better in an earlier version but there was a bug and I blamed the wrong thing at first, so I undid it, and forgot I could change it back after I found the real bug - latest version is better.

frontend/src/GameReportForm.tsx
chelseaSchmidt: Yes haha, god 🤦‍♀️

frontend/src/GameReportForm.tsx
chelseaSchmidt: Like it 💯

chelseaSchmidt:
@sirrus233 Thanks for all the good catches; the only surprising changes after your review might be in [20954c3](https://github.com/sirrus233/wotr-community-website/pull/15/commits/20954c37fc8de3084e90f31fa535ccbc87d53f9a) and [f68ba4b](https://github.com/sirrus233/wotr-community-website/pull/15/commits/f68ba4b71b5c4820a53b87873e11a9190bb85e95). Will go ahead and merge this to unblock follow-up work but if you have any questions/suggestions about those bits lmk and I'll take a look!

chelseaSchmidt:
@sirrus233 can you make the default `squash and merge` and add that auto-branch-deletion setting? Save me from myself

sirrus233:
"Such is of the course of deeds that move the wheels of the world: small hands do them because they must, while the eyes of the great are elsewhere."

sirrus233:
(Done.)

PR #16, opened 2024-12-13
Extract form state management into a custom hook

chelseaSchmidt:
To reduce file size. This makes `GameReportForm` responsible for the form's presentation and `useFormData` responsible for the form's state management. I copied the structure of the hook's return value (`[State, Meta, Helpers]`) from an npm package I like to use.

It's lost in the diff, but also renamed `controlledClearEffect` to `useControlledClearEffect` to be more React-y since it's a custom hook!

What do you think of the overall change? Does the splitting up make it hard to track the relationship between form state and presentation? These are some less nuclear alternatives I could do, which would still take good-sized chunks out of the `GameReportForm`:
- extract `initialFormData` into its own file
- extract `controlledClearEffect` and all its invocations into a smaller custom hook

### DIFF
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -1,5 +1,5 @@
 import "@fontsource/inter";
-import React, { useState, useEffect } from "react";
+import React from "react";
 import { FormControl, FormHelperText, List, Typography } from "@mui/joy";
 import Sheet from "@mui/joy/Sheet";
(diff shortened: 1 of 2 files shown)


sirrus233:
I think this is nice!!

:ship:

sirrus233 (approved):

chelseaSchmidt:
Yay will merge! Love having a custom hook or two just to feel fancy 🥳

PR #19, opened 2024-12-14
Haskell experiment

sirrus233:
This is a basic Haskell webserver! This change also removes all the Rust stuff, so you can ignore all those files.

The server itself has a single endpoint, but doesn't really serve anything interesting (just a 422 error).

Investigate to your heart's content! When in doubt, ask questions. I think there's a lot here that's not *incredibly* obvious. Make a special note of the `Types/App.hs` module, it is small but important.


chelseaSchmidt (approved):
BEAUTIFUL

PR #21, opened 2024-12-16
Return submitted report from submit api

sirrus233:
Now that the base work is done, gonna try to have these be somewhat more reasonable in size!

This changes the submitReport api to actually return the (processed) submitted report.

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -10,7 +10,7 @@ import Servant.Server (err422)
 import Types.Api (GameReport (..))
 import Types.App (AppM)
 import Types.DataField (PlayerId)
-import Types.Database (WriteProcessedGameReport (..))
+import Types.Database (ReadProcessedGameReport (..), WriteProcessedGameReport (..))
 import Validation (validateReport)
 
 processReport :: UTCTime -> PlayerId -> PlayerId -> GameReport -> WriteProcessedGameReport
@@ -40,14 +40,14 @@ processReport timestamp winner loser report =
       loserRatingAfter = 0 -- TODO
     }
 
-submitReportHandler :: GameReport -> AppM ()
+submitReportHandler :: GameReport -> AppM ReadProcessedGameReport
 submitReportHandler r = case validateReport r of
   Failure errors -> throwError $ err422 {errBody = show errors}
   Success report -> do
     winnerId <- insertPlayerIfNotExists report.winner
     loserId <- insertPlayerIfNotExists report.loser
     now <- liftIO getCurrentTime
-    insertGameReport . processReport now winnerId loserId $ report
+    insertGameReport $ processReport now winnerId loserId report
 
 server :: ServerT Api AppM
 server = submitReportHandler
--- a/backend/src/Database.hs
+++ b/backend/src/Database.hs
@@ -19,9 +19,9 @@ initializeDatabase = do
   env <- ask
   liftIO . withResource env.dbPool $ \conn -> do
     log env.logger "Initializing database"
-    tableCount <- query_ conn "SELECT COUNT(*) FROM sqlite_master WHERE type='table';" :: IO [Only Int]
+    tableCount <- readSingle $ query_ conn "SELECT COUNT(*) FROM sqlite_master WHERE type='table';" :: IO Int
     case tableCount of
-      [Only 0] -> do
+      0 -> do
         log env.logger "Creating tables"
         execute_ conn "PRAGMA foreign_keys = ON;"
         execute_
--- a/backend/src/Database.hs
+++ b/backend/src/Database.hs
@@ -1,10 +1,18 @@
 module Database where
 
+import Control.Exception (throwIO)
 import Data.Pool (withResource)
-import Database.SQLite.Simple (Only (..), execute, execute_, query, query_)
+import Database.SQLite.Simple (Only (..), execute_, query, query_)
 import Types.App (AppM, Env (..), log)
 import Types.DataField (PlayerId, PlayerName)
-import Types.Database (ReadPlayer (..), WritePlayer (..), WriteProcessedGameReport)
+import Types.Database (ReadPlayer (..), ReadProcessedGameReport (..), WritePlayer (..), WriteProcessedGameReport (..))
+
+data UnexpectedResultException = UnexpectedResultException deriving (Show)
+
+instance Exception UnexpectedResultException
+
+readSingle :: IO [Only r] -> IO r
+readSingle rs = rs >>= \case [Only r] -> pure r; _ -> throwIO UnexpectedResultException
 
 initializeDatabase :: ReaderT Env IO ()
 initializeDatabase = do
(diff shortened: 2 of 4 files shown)


backend/src/AppServer.hs
```diff
@@ -40,14 +40,14 @@ processReport timestamp winner loser report =
...
-submitReportHandler :: GameReport -> AppM ()
+submitReportHandler :: GameReport -> AppM ReadProcessedGameReport
 submitReportHandler r = case validateReport r of
   Failure errors -> throwError $ err422 {errBody = show errors}
   Success report -> do
     winnerId <- insertPlayerIfNotExists report.winner
     loserId <- insertPlayerIfNotExists report.loser
     now <- liftIO getCurrentTime
-    insertGameReport . processReport now winnerId loserId $ report
+    insertGameReport $ processReport now winnerId loserId report
```
sirrus233, line 50: The meaning of this line is identical, but I think putting the `$` there communicates the intent more clearly than the composition.

backend/src/Database.hs
```diff
@@ -1,10 +1,18 @@
...
 import Types.App (AppM, Env (..), log)
 import Types.DataField (PlayerId, PlayerName)
-import Types.Database (ReadPlayer (..), WritePlayer (..), WriteProcessedGameReport)
+import Types.Database (ReadPlayer (..), ReadProcessedGameReport (..), WritePlayer (..), WriteProcessedGameReport (..))
+
+data UnexpectedResultException = UnexpectedResultException deriving (Show)
+
+instance Exception UnexpectedResultException
+
+readSingle :: IO [Only r] -> IO r
```
sirrus233, line 14: Abstracting out a pattern here that I was starting to see in a few places.

Sometimes we want to read a single value from the database. Say, the `id` of a player. The query function always returns `IO [r]` where `r` is whatever your "row" type is. So if you're reading a `MyProductType` you get `IO [MyProductType]`, if you're reading only one field, you get an `IO [Only MyField]` (where Only is the single-element tuple), etc.

Unfortunately this is rather weakly typed (by the admittedly very high bar of our pseudo-dependently typed aspirations) and we ALWAYS get a list back, even if we are expecting a single result.*

So this helper is intended to be used when you know you are expecting one **and only one** thing, and you want to extract it, and throw an exception if anything else happens.

*Aside: what would be cool/hard for this library is type-level queries from which are returned some sized-list type who's size depends on the query :shrug: Guess we can't ask for the universe when some languages out there can't cogently express `this`.

backend/src/Validation.hs
```diff
@@ -113,8 +113,7 @@ validateStrongholds report
 
 validateReport :: GameReport -> Validation [ReportError] GameReport
 validateReport report =
-  report
```
sirrus233, line 116: Our discussion today made me realize the fmap is entirely unnecessary :smile: 
Don't need to fmap the report into the Validation... it's already in there.

backend/src/Database.hs
```diff
@@ -76,41 +85,69 @@ insertPlayerIfNotExists name = do
...
+
+    let nameQuery = "SELECT name FROM Players WHERE id = ?"
+    winner <- readSingle $ query conn nameQuery (Only report.winner)
+    loser <- readSingle $ query conn nameQuery (Only report.loser)
+
+    pure
+      ReadProcessedGameReport
+        { rid,
+          timestamp = report.timestamp,
+          winner,
```
sirrus233, line 132: Have you seen this pattern before? You're allowed to write this, instead of something like `winner = winner` as long as the field of the record has the same name as the variable in question. It's called a **record pun**. Neat convenience feature.

https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/record_puns.html

backend/src/Database.hs
chelseaSchmidt: Solid addition, this makes reading `tableCount`/`pid`/`winner`/`loser` very tidy looking! It's nice to have the extraction logic abstracted into one place, and throwing an exception in this case makes perfect sense to me.

> Guess we can't ask for the universe when some languages out there can't cogently express `this`

Dead 😂 💀 poor javascript out there just doing its best

backend/src/Database.hs
```diff
@@ -64,10 +72,11 @@ getPlayerByName :: PlayerName -> AppM (Maybe ReadPlayer)
 getPlayerByName name = do
   env <- ask
   liftIO . withResource env.dbPool $ \conn -> do
-    player <- query conn "SELECT * FROM Players WHERE name=(?)" (Only name)
-    case player of
-      [p] -> pure (Just p)
-      _ -> pure Nothing
+    result <- query conn "SELECT * FROM Players WHERE name = ?" (Only name)
```
chelseaSchmidt, line 75: What does changing `name=(?)` to `name = ?` do? Rest of the change makes sense ✔️

backend/src/Database.hs
chelseaSchmidt: Aah cool it's like property shorthand in JS! It's a moot point because `WriteProcessedGameReport` has properties that don't belong in `ReadProcessedGameReport`, but is it possible to `spread` too? Like would there be an equivalent to:
```
{
    rid,
    winner,
    loser,
    ...report,
}
```
This is probably just JS giving me crazy ideas 😬

chelseaSchmidt (approved):
🏅🏅🏅

backend/src/Database.hs
sirrus233: No difference, the parens are just optional there, so I took 'em out :)

chelseaSchmidt: 💯

backend/src/Database.hs
sirrus233: I don't think that's a crazy idea at all - but Haskell syntax doesn't support it. There are a few similar things you can do though...

Suppose we have the types:
```hs
data Foo = {a :: Int, b :: Int, c :: Int}
data Bar = {a :: Int, b :: Int, d :: Int}
```
You can decompose a value into its fields with a pattern match, so we could write something like:
```hs
-- Naive, no pattern match
mkBar :: Foo -> Bar
mkBar foo = Bar {a = foo.a, b = foo.b, d = 7}

-- Pattern match version. Matches *any* Foo, but doesn't assign a name to `c`
mkBar' :: Foo -> Bar
mkBar' (Foo {a b}) = Bar {a, b, d = 7}
```

There's also record *update* syntax, which is handy when you want to change only a subset of the fields:
```hs
-- Take a Bar and change it's `a` field to match the Foo's `a` field. Bar's `b` and `d` stay untouched
updateBar :: Foo -> Bar -> Bar
updateBar (Foo {a}) bar = bar {a}
```


EDIT!!!!
In the process of writing this response, I learned about https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/record_wildcards.html

This has to be enabled as a language extension, but looks very similar to record spreading when used in conjunction with the above tricks. I'm going to look into this to see if it can save us some verbosity!

chelseaSchmidt: Awesome find! Would be so cool if this pans out! (and the original pattern matching response is good learning for me too)

PR #22, opened 2024-12-16
Transaction support

sirrus233:
Change the monad context of the database helpers from `AppM` to just `IO`. This means we have to pass the connection explicitly, but makes it possible to group (what are now) IO actions into a single database transaction. We want our handlers to be able to do this, since they are frequently going to be making several queries which should either succeed or fail as a group.

Key new function is from `sqlite-simple` library:
> `withTransaction :: Connection -> IO a -> IO a`
Run an IO action inside a SQL transaction started with BEGIN TRANSACTION. If the action throws any kind of an exception, the transaction will be rolled back with ROLLBACK TRANSACTION. Otherwise the results are committed with COMMIT TRANSACTION.

Note: I suspect (but did not try) that we could keep the database functions inside `AppM` if we changed them from returning "a value embedded in `AppM`" to returning "an `IO` action that yields a value, embedded in `AppM`. So the database functions return the IO actions, and those get composed into a transaction in the handler layer. This sounds a little annoying to deal with even if it works, so I advocate for doing it this way until we have a good reason to need more.

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -1,14 +1,16 @@
 module AppServer where
 
 import Api (Api)
+import Data.Pool (withResource)
 import Data.Time (UTCTime)
 import Data.Time.Clock (getCurrentTime)
 import Data.Validation (Validation (..))
 import Database (insertGameReport, insertPlayerIfNotExists)
+import Database.SQLite.Simple (withTransaction)
 import Servant (ServerError (errBody), ServerT, throwError)
 import Servant.Server (err422)
 import Types.Api (GameReport (..))
-import Types.App (AppM)
+import Types.App (AppM, Env (..))
 import Types.DataField (PlayerId)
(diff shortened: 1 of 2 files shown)


chelseaSchmidt (approved):
Passing the connection as an explicit arg is definitely worth it just for the code readability of having the `liftIO . withResource env.dbPool $ \conn -> do` bit happen in one spot. And am I understanding right that now there's one grouped SQL query sent to the database where before there were multiple sequential queries sent separately to the database? That's a sweet improvement!

`withTransaction` ftw 💪

sirrus233:
@chelseaSchmidt yeah in an SQL transaction, multiple queries are executed (still sequentially) but no data is actually modified unless all queries succeed.

There's a stronger form of transaction where the database is locked when the transaction begins, and then unlocked only once the transaction is complete. I'm not doing that here, but we could if it's necessary.

PR #25, opened 2024-12-17
Submitting a report modifies Elo

sirrus233:
This about does it for our basic implementation of submitting a report.

* We receive the report
* Insert players into the DB if necessary
* Insert the report into the DB
* Insert a row representing ratings diffs into the DB

I ended up walking away from storing ratings in the report table directly - kind of a headache. Check out the new schema, as well as the API's return type - see what you think!

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -1,38 +1,98 @@
 module AppServer where
 
 import Api (Api)
+import Data.IntMap.Strict qualified as Map
 import Data.Pool (withResource)
-import Data.Time (UTCTime)
 import Data.Time.Clock (getCurrentTime)
 import Data.Validation (Validation (..))
-import Database (insertGameReport, insertPlayerIfNotExists)
+import Database (getLatestRating, insertGameReport, insertPlayerIfNotExists, insertRatingChange)
 import Database.SQLite.Simple (withTransaction)
 import Servant (ServerError (errBody), ServerT, throwError)
 import Servant.Server (err422)
-import Types.Api (GameReport (..))
+import Types.Api (GameReport (..), SubmitGameReportResponse (..))
 import Types.App (AppM, Env (..))
-import Types.DataField (PlayerId)
-import Types.Database (ReadProcessedGameReport (..), WriteProcessedGameReport (..))
+import Types.DataField (Rating, Side (..))
+import Types.Database (ReadProcessedGameReport (..), WriteProcessedGameReport (..), WriteRatingChange (..))
 import Validation (validateReport)
 
-processReport :: UTCTime -> PlayerId -> PlayerId -> GameReport -> WriteProcessedGameReport
-processReport timestamp winnerId loserId (GameReport {..}) =
-  WriteProcessedGameReport
-    { winnerRatingAfter = 0, -- TODO
-      loserRatingAfter = 0, -- TODO
-      ..
-    }
+defaultRating :: Rating
+defaultRating = 500
 
-submitReportHandler :: GameReport -> AppM ReadProcessedGameReport
+ratingThresholds :: IntMap (Rating, Rating)
+ratingThresholds =
+  fromList
+    [ (10, (16, 16)),
+      (33, (15, 17)),
+      (56, (14, 18)),
+      (79, (13, 19)),
+      (102, (12, 20)),
+      (126, (11, 21)),
+      (151, (10, 22)),
+      (178, (9, 23)),
+      (207, (8, 24)),
+      (236, (7, 25)),
+      (270, (6, 26)),
+      (308, (5, 27)),
+      (352, (4, 28)),
+      (409, (3, 29)),
+      (499, (2, 30))
+    ]
+
+maxThreshold :: (Rating, Rating)
+maxThreshold = (1, 31)
+
+ratingAdjustment :: Rating -> Rating -> Rating
+ratingAdjustment winner loser
+  | winner >= loser = smallAdjust
+  | otherwise = bigAdjust
+  where
+    diff = abs (winner - loser)
+    (smallAdjust, bigAdjust) = maybe maxThreshold snd (Map.lookupGE diff ratingThresholds)
+
+submitReportHandler :: GameReport -> AppM SubmitGameReportResponse
 submitReportHandler r = case validateReport r of
   Failure errors -> throwError $ err422 {errBody = show errors}
-  Success report -> do
+  Success (GameReport {..}) -> do
     env <- ask
     liftIO . withResource env.dbPool $ \conn -> withTransaction conn $ do
-      winnerId <- insertPlayerIfNotExists conn report.winner
-      loserId <- insertPlayerIfNotExists conn report.loser
-      now <- liftIO getCurrentTime
-      insertGameReport conn $ processReport now winnerId loserId report
+      timestamp <- liftIO getCurrentTime
+      winnerId <- insertPlayerIfNotExists conn winner
+      loserId <- insertPlayerIfNotExists conn loser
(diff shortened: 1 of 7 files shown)


backend/src/Database.hs
```diff
@@ -5,22 +5,31 @@ import Data.Pool (withResource)
...
+readOne :: IO [r] -> IO r
+readOne rs = rs >>= \case [r] -> pure r; _ -> throwIO UnexpectedResultException
 
 initializeDatabase :: ReaderT Env IO ()
 initializeDatabase = do
   env <- ask
   liftIO . withResource env.dbPool $ \conn -> do
     log env.logger "Initializing database"
-    tableCount <- readSingle $ query_ conn "SELECT COUNT(*) FROM sqlite_master WHERE type='table';" :: IO Int
+    tableCount <- fromOnly <$> readOne (query_ conn "SELECT COUNT(*) FROM sqlite_master WHERE type='table';") :: IO Int
```
chelseaSchmidt, line 32: `fromOnly`, cool!

backend/src/Database.hs
```diff
@@ -49,11 +58,9 @@ initializeDatabase = do
           \strongholds TEXT NOT NULL,\
           \interestRating INTEGER NOT NULL,\
           \comments TEXT,\
-          \winnerRatingAfter INTEGER NOT NULL,\
-          \loserRatingAfter INTEGER NOT NULL,\
```
chelseaSchmidt, line 53: > I ended up walking away from storing ratings in the report table directly - kind of a headache

This makes sense, now that you mention it, these do seem redundant when you can just find them in the Ratings table, plus you added an index so there's no real drawback. I can't remember what the original rationale was. But anyway was the headache because `insertRatingChange` had to happen after `insertGameReport`?

backend/src/Database.hs
```diff
@@ -63,36 +70,65 @@ initializeDatabase = do
...
+        execute_
+          conn
+          "CREATE TABLE Ratings (\
+          \id INTEGER NOT NULL,\
+          \playerId INTEGER NOT NULL,\
+          \side TEXT NOT NULL,\
+          \timestamp INTEGER NOT NULL,\
+          \reportId INTEGER NOT NULL,\
+          \ratingBefore INTEGER NOT NULL,\
+          \ratingAfter INTEGER NOT NULL,\
```
chelseaSchmidt, line 82: You were debating whether to include one or both before, did this just end up being the simpler choice to start with, and can get fancy later?

backend/src/Database.hs
```diff
@@ -63,36 +70,65 @@ initializeDatabase = do
...
       _ -> log env.logger "Database already initialized"
 
 getPlayerByName :: SQL.Connection -> PlayerName -> IO (Maybe ReadPlayer)
-getPlayerByName conn name = do
-  result <- query conn "SELECT * FROM Players WHERE name = ?" (Only name)
-  case result of
-    [] -> pure Nothing
-    [player] -> pure (Just player)
-    _ -> throwIO UnexpectedResultException
+getPlayerByName conn name = readZeroOrOne $ query conn "SELECT * FROM Players WHERE name = ?" (Only name)
```
chelseaSchmidt, line 92: Love how much shorter the extra helper function made this

backend/src/AppServer.hs
```diff
@@ -1,38 +1,98 @@
...
+maxThreshold :: (Rating, Rating)
+maxThreshold = (1, 31)
+
+ratingAdjustment :: Rating -> Rating -> Rating
+ratingAdjustment winner loser
+  | winner >= loser = smallAdjust
+  | otherwise = bigAdjust
+  where
+    diff = abs (winner - loser)
+    (smallAdjust, bigAdjust) = maybe maxThreshold snd (Map.lookupGE diff ratingThresholds)
```
chelseaSchmidt, line 50: It says a lot about your code readability that this made sense to me! Am I reading it right that the logic is:

1. Get the difference between their ratings and look up the corresponding `(smallAdjust, bigAdjust)` in the `ratingThresholds` map.
2. If the difference is higher than 499, use `(1, 31)` as the adjustment options
3. If the winner was previously rated below the user, apply the `bigAdjust`; otherwise the `smallAdjust`

chelseaSchmidt (approved):
Wooo this is a big step! Sorry for delay in reviewing, was MIA yesterday and couldn't get my Haskell fix. This is very nice 👏

backend/src/AppServer.hs
sirrus233: Yep! Exactly what you said. Beating someone higher ranked than you gets you more points.

backend/src/Database.hs
sirrus233: This is from the sqlite-simple library, and is a very common pattern in haskell. Since we often have values wrapped in some kind of container type (`Only` in this case, but we do it all the time) -- there's two main ways to "extract" the internal value. One is pattern matching, but the other is to provide a utility function that extracts the value for you.

sirrus233: Also, not sure if it's obvious here, but the fmap is to apply the fromOnly inside the IO returned by readOne

backend/src/Database.hs
sirrus233: Originally we were talking about not having a ratings table because it's 1:1 with game reports. 

But I found that storing ratings on the game report got awkward to query, because a report row only stores the Side played by the winner. So if you want, say, "Chelsea's latest Shadow rating", you need to query for "the last game report with Side Shadow that Chelsea won and then get winnerRatingAfter, OR the last game report with Side Free that Chelsea lost and then get loserRstingAfter" and this was starting to make me dizzy. 

So I think this keeps things a lot more clear, at least in my head.

sirrus233: Now we store two rating rows per report row. One per player, each annotated with Side

backend/src/Database.hs
sirrus233: I suspect/realized `ratingBefore` (which is currently unnecessary) will make historical ratings adjustments much easier to do if/when those become necessary

sirrus233: Because there will be no need to "find the last game played prior to the one you're adjusting"

backend/src/Database.hs
sirrus233: I realized a shortcoming in the sqlite-simple library while writing this. Because the return type of `query` is so loose (`[r] for any r`) type inference is pretty useless and the haskell side of things has no choice but to believe whatever you tell it is inside the list. This can manifest in runtime errors if that's not actually what's in the list after the query is performed. 

This is the cost of getting to write very simple untyped SQL statements. Probably worth it for now, but it's got me interested to see what other SQL library options are available that might be more complicated to use, but more type safe

sirrus233 (commented):
Feedback addressed

backend/src/Database.hs
sirrus233: Yeahh https://hackage.haskell.org/package/esqueleto

backend/src/Database.hs
chelseaSchmidt: Still takes my brain an extra step to read `<$>` as `fmap` instead of thinking it's the applicative one `<*>` (idk, it's a mental block) so in that sense it was not 😂  but I did pick up on it being needed for lifting `fromOnly` into the IO! Baby steps

backend/src/Database.hs
chelseaSchmidt: Gotcha, it was just a leftover from before the Ratings table decision, now I'm caught up! Dizzy from following the old way as well haha

backend/src/Database.hs
chelseaSchmidt: Good point 💯

PR #26, opened 2024-12-19
Mask null in number input form control

chelseaSchmidt:
To fix a console error `Failed prop type: The prop value is marked as required in ForwardRef(StableOption), but its value is null`

This might be why you made the initial value `-1` originally! I like `null` remaining the true representation of "N/A" in form state, so what do you think of this as a middle ground?

### DIFF
--- a/frontend/src/SelectNumericOptionInput.tsx
+++ b/frontend/src/SelectNumericOptionInput.tsx
@@ -2,6 +2,8 @@ import React from "react";
 import SelectOptionInput from "./SelectOptionInput";
 import { INFINITE } from "./constants";
 
+const NULL_MASK = -1;
+
 interface SelectNumericOptionInputProps {
     start: number;
     end: number;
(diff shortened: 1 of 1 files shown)


frontend/src/SelectNumericOptionInput.tsx
```diff
@@ -32,16 +34,20 @@ export default function SelectNumericOptionInput({
...
+                if (value === NULL_MASK) {
                     return "";
                 } else if (value === INFINITE) {
                     return "Unlimited!";
                 } else {
                     return String(value);
                 }
             }}
-            onChange={onChange}
+            onChange={(val) => onChange(unmaskNull(val))}
```
chelseaSchmidt, line 45: Wanted to pass in `unmaskNull . onChange` 🥲

sirrus233 (approved):
LGTM!

Another possibility to consider: since there's already going to need to be a mapping layer between the UI values and the actual http payload, we could do the -1 -> null mapping there instead 🤷

Not necessarily advocating for this, but it's an option

chelseaSchmidt:
True, didn't consider that! I waffled for a second and leaning toward it being isolated to the local form control to keep "N/A" more consistent in initial values. Will merge for now and possibly change my mind later 🫠

PR #27, opened 2024-12-19
Use FormControl wrapper only around single form controls

chelseaSchmidt:
Bandaid for the console error `Joy: A FormControl can contain only one control component (Autocomplete | Input | Textarea | Select | RadioGroup) You should not mix those components inside a single FormControl instance`

I don't love the boolean toggle I did, but I didn't want to to change the implementation of `MultiOptionInput` to use `Select` somehow and be a dropdown multiselect instead of the pretty buttons, and I didn't want a second version of `GameReportFormElement` that might get out of sync in presentation style with the original one, so it's what I landed on 🤷‍♀️ Is there a secret fourth option that would be better?

edit: I guess one more option is to discard the use of `FormControl` completely, but I like the styling it gives to the non-multi-select inputs when there's an error

### DIFF
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -293,10 +293,9 @@ function GameReportForm() {
             <GameReportFormElement
                 label={"What strongholds were captured by Shadow?"}
                 error={formData.strongholds.error}
+                hasSingleControl={false}
             >
-                <VictoryPoints
-                    strongholds={formData.strongholds.value}
-                ></VictoryPoints>
+                <VictoryPoints strongholds={formData.strongholds.value} />
                 <MultiOptionInput
                     values={strongholds.slice()}
                     current={formData.strongholds.value}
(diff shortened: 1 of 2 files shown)


sirrus233:
So, I'm unclear on why this should or shouldn't be allowed - do we know what the warning is actually warning us of?

It seems that in some cases, the warning is a known bad warning and the prevailing advice is just "ignore it" (JavaScript culture I guess) https://github.com/mui/material-ui/issues/37764

Is this one of those cases? Does this PR address a real issue, or just clear a warning?

chelseaSchmidt:
I did some element inspecting and it's probably mostly accessibility. `FormControl` is causing each of the checkboxes to be assigned the same `id` and getting two `labels` assigned via their button label's `for` and the `FormLabel`'s `for`.

I slightly favor keeping accessibility because ✨ **rules** ✨, but at the same time I doubt anyone is playing WotR with a screen reader 😂 so if you vote against I will throw this away!

sirrus233:
No I'm good with it! Makes sense to me 😃

PR #28, opened 2024-12-21
Map form data to a schematically valid HTTP payload

chelseaSchmidt:
⚠️ *Does not include mapping of stronghold names to unspaced values* ⚠️

This does a bare minimum transformation of the form data into the schema expected by the server and allows a successful form submission as we saw earlier, hooray! Code-wise though, I was fighting for my life with types on this one and I don't know if what I landed on was the best solution? Will add more details in-line.

### DIFF
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -62,6 +62,28 @@ export const optionalFields = [
     "comment",
 ] as const;
 
+export const payloadFields = [
+    "winner",
+    "loser",
+    "side",
+    "victory",
+    "match",
+    "competition",
+    "league",
+    "expansions",
+    "treebeard",
+    "actionTokens",
+    "dwarvenRings",
+    "turns",
+    "corruption",
+    "mordor",
+    "initialEyes",
+    "aragornTurn",
+    "strongholds",
+    "interestRating",
+    "comment",
+] as const;
+
 export enum ErrorMessage {
     Required = "Required",
     OnSubmit = "Could not submit, please resolve errors",
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -49,6 +49,16 @@ export const cities: (typeof strongholds)[number][] = [
     "IronHills",
 ];
 
+export const optionalFields = [
+    "league",
+    "treebeard",
+    "actionTokens",
+    "dwarvenRings",
+    "mordor",
+    "aragornTurn",
+    "comment",
+] as const;
+
 export enum ErrorMessage {
     Required = "Required",
     OnSubmit = "Could not submit, please resolve errors",
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -63,25 +63,16 @@ export const optionalFields = [
 ] as const;
 
 export const payloadFields = [
+    ...optionalFields,
     "winner",
     "loser",
     "side",
     "victory",
     "match",
-    "competition",
-    "league",
-    "expansions",
-    "treebeard",
-    "actionTokens",
-    "dwarvenRings",
     "turns",
     "corruption",
-    "mordor",
     "initialEyes",
-    "aragornTurn",
-    "strongholds",
     "interestRating",
-    "comment",
 ] as const;
 
 export enum ErrorMessage {
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -50,12 +50,15 @@ export const cities: (typeof strongholds)[number][] = [
 ];
 
 export const optionalFields = [
+    "competition",
     "league",
+    "expansions",
     "treebeard",
     "actionTokens",
     "dwarvenRings",
     "mordor",
     "aragornTurn",
+    "strongholds",
     "comment",
 ] as const;
 
--- a/frontend/src/hooks/useFormData/index.tsx
+++ b/frontend/src/hooks/useFormData/index.tsx
@@ -37,12 +37,13 @@ const useFormData = (): [FormData, Meta, Helpers] => {
     const validateField = <K extends keyof FormData>(field: K) => {
         return () => {
             setFormData((prevData) => {
-                const fieldError = prevData[field].validate();
-                const isMissing = isFieldMissing(field, prevData);
                 const error: FieldError =
-                    fieldError || (isMissing && ErrorMessage.Required) || null;
+                    prevData[field].validate() ||
+                    (isFieldMissing(field, prevData) &&
+                        ErrorMessage.Required) ||
+                    null;
 
-                return fieldError || prevData[field].error || isMissing
+                return error || prevData[field].error
                     ? {
                           ...prevData,
                           [field]: { ...prevData[field], error },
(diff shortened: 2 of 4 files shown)


frontend/src/hooks/useFormData/index.tsx
```diff
@@ -43,43 +52,48 @@ const useFormData = (): [FormData, Meta, Helpers] => {
...
         );
 
-        if (stateUpdates.length) {
+        if (noErrorsDetected(stateUpdates, formData)) {
+            return formData;
+        } else {
             stateUpdates.forEach((update) => update());
             return ErrorMessage.OnSubmit;
-        } else {
-            return null;
```
chelseaSchmidt, line 67: First issue with this PR is here.

Background, I inverted `validateForm` to return the equivalent of an Either (just now realizing that when you were talking about doing a Maybe/Either in an earlier PR, you just meant `X | Y`, not actually implementing a type called `Maybe`/`Either`? 🤦‍♀️ ). Anyway, returns either: valid form data, or an error message.

I did that to help with a second change, which is returning `formData` after passing it through  `noErrorsDetected`, which is a helper function that coerces `formData` from type `FormData` into`ValidFormData`. `ValidFormData` is the same thing, except it does not allow `null` for required fields.

This lets me transform further into the `GameReportPayload`, which requires certain previously nullable values to now be non-nullable values.

Basically my problem with it: it opens us up to a `400` error from the server if the validation were to not properly confirm non-null values, since I'm just blanket telling Typescript to believe me.

frontend/src/hooks/useFormData/index.tsx
```diff
@@ -151,19 +165,52 @@ const useFormData = (): [FormData, Meta, Helpers] => {
...
+        actionTokens: formData.actionTokens.value,
+        dwarvenRings: formData.dwarvenRings.value,
+        turns: formData.turns.value,
+        corruption: formData.corruption.value,
+        mordor: formData.mordor.value,
+        initialEyes: formData.initialEyes.value,
+        aragornTurn: formData.aragornTurn.value,
+        strongholds: formData.strongholds.value,
+        interestRating: formData.interestRating.value,
+        comment: formData.comment.value,
```
chelseaSchmidt, line 212: Second issue: this block is gross and I want the Haskell `..` 😭  I tried to create a helper function like below to at least do the transformation to values first, which was only an initial step, and couldn't get even get that working:
```
type ReducedFormData = { [K in keyof FormData]: FormData[K]["value"] };

function transformValues(
    formData: FormData,
    transform: <K extends keyof FormData>(
        key: K,
        obj: FormData
    ) => FormData[K]["value"]
): ReducedFormData {
    return objectKeys(formData).reduce((accum, key) => {
        if (key in formData) {
            accum[key] = transform(key, formData);
        }
        return accum;
    }, {} as ReducedFormData);
}
```
Any ideas? And/or is it worth it to keep trying? The TS error is on `accum[key] = transform(key, formData);`, and it's:
```
Type 'string | number | boolean | ("League" | "Tournament")[] | ("LoME" | "WoME" | "KoME" | "Cities" | "Fate of Erebor" | "Treebeard")[] | ("Rivendell" | "Grey Havens" | "Helm's Deep" | ... 10 more ... | "Iron Hills (Fate of Erebor expansion only)")[] | null' is not assignable to type 'never'.
  Type 'null' is not assignable to type 'never'
```

frontend/src/hooks/useFormData/initialFormData.tsx
```diff
@@ -1,41 +1,30 @@
...
+import { FormData } from "../../types";
 
 const initialFormData: FormData = {
     winner: {
         value: null,
         error: null,
-        validate: function _() {
-            return detectMissingInput(this.value);
-        },
+        validate: alwaysValid,
```
chelseaSchmidt, line 7: I moved responsibility for checking required fields into the `validateField` and `validateForm` helpers, because null/non-null validation is specifically neeeded to ensure the type coercion from `FormData` to `ValidFormData` isn't a bad idea, so I wanted all the logic to at least be closer together.

So all the `validate` functions on individual fields are placeholders now. If they end up staying that way (haven't thought about best place for dependent field validation yet), will wipe them out!

chelseaSchmidt:
Done commenting about the things I don't like, up for review!

sirrus233 (approved):
Approving, so as not to block - but left some of my thoughts. Feel free to hmu if you wanna chat about any of it!

frontend/src/hooks/useFormData/index.tsx
chelseaSchmidt: This is a sensible way of laying out the options, I think I will stick with the trade of more lines of code in exchange for keeping type safety! But agnostic of that you're right that `GameReportPayload` would be better as a mapped type, nice. Added that: [2620b0b](https://github.com/sirrus233/wotr-community-website/pull/28/commits/2620b0b25705e1e43ba64536e8ce07005ac31d8e)

frontend/src/hooks/useFormData/index.tsx
chelseaSchmidt: > a pattern called a Type Guard, where you narrow a type based on runtime checking of the value semantics. There isn't really anything wrong with this approach. It does have the downside you've called out, which is that your type-checking is vulnerable now to runtime logic, but that doesn't make it an invalid technique

TIL, I didn't know there was a name for this, that's really cool and I feel better about what I did!

> If you do keep the type guard, I would perhaps consider ditching the helper function for clarity. It's probably simpler to parse this as `if (stateUpdates.length) return formData as ValidFormData`

This is much simpler and easier to see clearly what's going on when it's not moved off into a function, updated ✔️ 

> If you really hate the type guard, then you need to construct the `ValidFormData` explicitly while performing the appropriate `null` checking

Now that you explained what the pattern is I don't hate it anymore! I feel like I'm making the opposite decision here to the mapping from `ValidFormData` to `GameReportPayload`, but in that case the tradeoff is only more lines of code, while in this case it would mean more lines of code AND doing null checking twice, and it just doesn't feel worth it.

frontend/src/constants.ts
```diff
@@ -43,6 +43,19 @@ export const cities: (typeof strongholds)[number][] = [
     "Iron Hills (Fate of Erebor expansion only)",
 ];
 
+export const optionalFields = [
```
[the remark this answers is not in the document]
chelseaSchmidt: Ooh smart! I didn't look at this comment before adding another const array for `payloadFields` in one of the recent commits. I'm looking at this strategy for eliminating both those arrays now 👀

frontend/src/hooks/useFormData/index.tsx
```diff
@@ -43,43 +52,48 @@ const useFormData = (): [FormData, Meta, Helpers] => {
...
+    const validateForm = (): ValidFormData | ErrorMessage.OnSubmit => {
         const stateUpdates = objectKeys(formData).reduce<(() => void)[]>(
             (updates, field) => {
                 const fieldError = formData[field].validate();
-                if (fieldError) {
+                const isMissing = isFieldMissing(field, formData);
+                const error: FieldError =
+                    fieldError || (isMissing && ErrorMessage.Required) || null;
+
+                if (fieldError || isMissing) {
```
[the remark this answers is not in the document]
chelseaSchmidt: Oops yeah good catch 😬

frontend/src/constants.ts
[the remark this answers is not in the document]
chelseaSchmidt: Discorded about this, abandoned, am sad

chelseaSchmidt:
@sirrus233 All feedback addressed (thank you so much), but there were substantial changes; do you want to look again or should I just merge?

sirrus233 (approved):

frontend/src/constants.ts
```diff
@@ -49,6 +49,41 @@ export const cities: (typeof strongholds)[number][] = [
     "IronHills",
 ];
 
+export const optionalFields = [
```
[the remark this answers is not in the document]
chelseaSchmidt: Yes! And I went to copy how it's done with `cities` and `strongholds`, but found that using the spread operator makes it DRY too! Extra nice ✨

PR #29, opened 2024-12-21
Map stronghold and expansion values to display labels

chelseaSchmidt:
The title! Tested it on the other branch and the server accepts all stronghold values after adding this mapping.

### DIFF
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -12,6 +12,7 @@ import {
     strongholds,
     victoryTypes,
 } from "./constants";
+import { Expansion, Stronghold } from "./types";
 import useFormData from "./hooks/useFormData";
 import GameReportFormElement from "./GameReportFormElement";
 import MultiOptionInput from "./MultiOptionInput";
@@ -147,6 +148,7 @@ function GameReportForm() {
                         current={formData.expansions.value}
                         onChange={handleInputChange("expansions")}
                         validate={validateField("expansions")}
+                        getLabel={getExpansionLabel}
                     />
                 </GameReportFormElement>
             )}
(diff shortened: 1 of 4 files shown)


frontend/src/GameReportForm.tsx
```diff
@@ -336,3 +339,43 @@ function GameReportForm() {
 }
 
 export default GameReportForm;
+
+function getStrongholdLabel(stronghold: Stronghold): string {
+    switch (stronghold) {
+        case "Rivendell":
```
sirrus233, line 345: What is this cursed syntax?? 🤣 Does the input get returned by default if the body of the `case` is empty??

sirrus233: Oh, nevermind, I see..

okay.

sirrus233: 🙃

sirrus233 (approved):
Great stuff! 👍 

Needs to be done for League as well, I believe. Here's the backend type:
```hs
data League = GeneralLeague | LoMELeague | WoMELeague | SuperLeague | TTSLeague deriving (Eq, Generic, Read, Show)
```

frontend/src/GameReportForm.tsx
chelseaSchmidt: FINALLY I get to use this random language feature for its intended purpose instead of it being a disastrous mistake 😂 it's so cursed

chelseaSchmidt:
> Needs to be done for League as well, I believe. Here's the backend type:

Missed that, thank you! Added and tested ✔️ 

FYI the league label mapping required a change to prop types in [SingleOptionInput](https://github.com/sirrus233/wotr-community-website/pull/29/files#diff-33d2da3eca2271214033a4f9f97c79430062e7f985bded623acf7eed4301861a). `values` - the radio options - should never include null, even though `current` should be nullable for representing unselected. This turns out to be applicable for all the radio option controls, but wasn't apparent before until there was a label mapping, because the alternative was adding a `null` case to the league label switch statement which should not be a thing. Will merge but wanted to give a heads up since that might look unexpected!

PR #30, opened 2024-12-23
Why not re-write the server

sirrus233:
I accidentally the server.

The big change here is from using the `sqlite-simple` library to the much more complex/powerful `persistent` library, alongside the companion library `esqueleto` (skeleton in Portuguese, I'm told). I was getting frustrated with how error prone the basic types provided by `sqlite-simple` were - things I would easily accept in any other language, but if we're going to be in Haskell, let's have nice things.

`persistent` is a general library for handling schematized storage in a type-safe way, it has many backends for different kinds of databases (we're still using the sqlite backend), handles schema migrations automagically, and all sorts of nice stuff. `esqueleto` allows for very powerful type-safe queries against the `persistent` database types.

I've had to touch a lot of code to make the change, but I'm hoping the end result will be worth it (or at least make us happier). Please take a look and see what you think - I've left a few TODO's for fast follow-up, but I think this is good for a review pass.

### DIFF
--- a/backend/app/Main.hs
+++ b/backend/app/Main.hs
@@ -7,13 +7,13 @@ import Database (initializeDatabase)
 import Database.Redis (ConnectInfo, connect, defaultConnectInfo, disconnect)
 import Database.SQLite.Simple (close, open)
 import Network.Wai.Handler.Warp (run)
+import Network.Wai.Middleware.Cors (CorsResourcePolicy (..), cors)
 import Servant (Application, hoistServer)
 import Servant.Server (serve)
 import System.Directory (createDirectoryIfMissing)
 import System.FilePath (takeDirectory)
 import System.Log.FastLogger (LogType, LogType' (..), defaultBufSize, newTimeCache, newTimedFastLogger, simpleTimeFormat)
 import Types.App (Env (..), log, nt)
-import Network.Wai.Middleware.Cors (cors, CorsResourcePolicy (..))
 
 databaseFile :: FilePath
 databaseFile = "data/db.sqlite"
@@ -24,17 +24,23 @@ redisConfig = defaultConnectInfo
 logType :: LogType
 logType = LogStdout defaultBufSize
 
+corsMiddleware :: Application -> Application
+corsMiddleware = cors . const $ Just policy
+  where
+    policy =
+      CorsResourcePolicy
+        { corsOrigins = Just (["http://127.0.0.1:3000"], True),
+          corsMethods = ["POST"],
+          corsRequestHeaders = ["content-type"],
+          corsExposedHeaders = Nothing,
+          corsMaxAge = Nothing,
+          corsVaryOrigin = True,
+          corsRequireOrigin = False,
+          corsIgnoreFailures = True
+        }
+
 app :: Env -> Application
-app env = cors (const $ Just CorsResourcePolicy {
-  corsOrigins = Just (["http://127.0.0.1:3000"], True),
-  corsMethods = ["POST"],
-  corsRequestHeaders = ["content-type"],
-  corsExposedHeaders = Nothing,
-  corsMaxAge = Nothing,
-  corsVaryOrigin = True,
-  corsRequireOrigin = False,
-  corsIgnoreFailures = True
-  }) . serve api . hoistServer api (nt env) $ server
+app env = corsMiddleware . serve api . hoistServer api (nt env) $ server
 
 main :: IO ()
 main = do
--- a/backend/app/Main.hs
+++ b/backend/app/Main.hs
@@ -47,6 +47,7 @@ main :: IO ()
 main = do
   createDirectoryIfMissing True . takeDirectory $ databaseFile
 
+  -- TODO Fix up debug logging
   dbPool <- runStdoutLoggingT $ createSqlitePoolWithConfig (toText databaseFile) defaultConnectionPoolConfig
   redisPool <- connect redisConfig
   timeCache <- newTimeCache simpleTimeFormat
@@ -54,6 +55,7 @@ main = do
 
   let env = Env {dbPool, redisPool, logger}
 
+  -- TODO Disable/handle auto-migration
   runSqlPool (runMigration migrateAll) dbPool
 
   log logger "Starting server"
--- a/backend/app/Main.hs
+++ b/backend/app/Main.hs
@@ -2,10 +2,11 @@ module Main where
 
 import Api (api)
 import AppServer (server)
-import Data.Pool (defaultPoolConfig, destroyAllResources, newPool)
-import Database (initializeDatabase)
+import Control.Monad.Logger (runStdoutLoggingT)
+import Data.Pool (destroyAllResources)
+import Database.Esqueleto.Experimental (defaultConnectionPoolConfig, runMigration, runSqlPool)
+import Database.Persist.Sqlite (createSqlitePoolWithConfig)
 import Database.Redis (ConnectInfo, connect, defaultConnectInfo, disconnect)
-import Database.SQLite.Simple (close, open)
 import Network.Wai.Handler.Warp (run)
 import Network.Wai.Middleware.Cors (CorsResourcePolicy (..), cors)
 import Servant (Application, hoistServer)
@@ -14,6 +15,7 @@ import System.Directory (createDirectoryIfMissing)
 import System.FilePath (takeDirectory)
 import System.Log.FastLogger (LogType, LogType' (..), defaultBufSize, newTimeCache, newTimedFastLogger, simpleTimeFormat)
 import Types.App (Env (..), log, nt)
+import Types.Database (migrateAll)
 
 databaseFile :: FilePath
 databaseFile = "data/db.sqlite"
@@ -25,7 +27,7 @@ logType :: LogType
 logType = LogStdout defaultBufSize
 
 corsMiddleware :: Application -> Application
-corsMiddleware = cors . const $ Just policy
+corsMiddleware = cors $ const $ Just policy
   where
     policy =
       CorsResourcePolicy
@@ -46,14 +48,14 @@ main :: IO ()
 main = do
   createDirectoryIfMissing True . takeDirectory $ databaseFile
 
-  dbPool <- newPool $ defaultPoolConfig (open databaseFile) close 30 10
+  dbPool <- runStdoutLoggingT $ createSqlitePoolWithConfig (toText databaseFile) defaultConnectionPoolConfig
   redisPool <- connect redisConfig
   timeCache <- newTimeCache simpleTimeFormat
   (logger, rmLogger) <- newTimedFastLogger timeCache logType
 
   let env = Env {dbPool, redisPool, logger}
 
-  runReaderT initializeDatabase env
+  runSqlPool (runMigration migrateAll) dbPool
 
   log logger "Starting server"
   run 8081 . app $ env
--- a/backend/app/Main.hs
+++ b/backend/app/Main.hs
@@ -3,10 +3,9 @@ module Main where
 import Api (api)
 import AppServer (server)
 import Control.Monad.Logger (runStdoutLoggingT)
-import Data.Pool (destroyAllResources)
 import Database.Esqueleto.Experimental (defaultConnectionPoolConfig, runMigration, runSqlPool)
 import Database.Persist.Sqlite (createSqlitePoolWithConfig)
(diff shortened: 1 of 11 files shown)


backend/app/Main.hs
```diff
@@ -24,35 +25,38 @@ redisConfig = defaultConnectInfo
 logType :: LogType
 logType = LogStdout defaultBufSize
 
+corsMiddleware :: Application -> Application
```
sirrus233, line 28: Just moved this into its own function for clarity, no changes.

backend/app/Main.hs
```diff
@@ -24,35 +25,38 @@ redisConfig = defaultConnectInfo
...
-  corsRequireOrigin = False,
-  corsIgnoreFailures = True
-  }) . serve api . hoistServer api (nt env) $ server
+app env = corsMiddleware . serve api . hoistServer api (nt env) $ server
 
 main :: IO ()
 main = do
   createDirectoryIfMissing True . takeDirectory $ databaseFile
 
-  dbPool <- newPool $ defaultPoolConfig (open databaseFile) close 30 10
```
sirrus233, line 43: The `persistent` backend manages its own resource pool, so no need to use the `resource-pool` api directly anymore (in fact I've removed it as an explicit dependency of the project, since we don't use it directly).

backend/app/Main.hs
```diff
@@ -24,35 +25,38 @@ redisConfig = defaultConnectInfo
...
+  dbPool <- runStdoutLoggingT $ createSqlitePoolWithConfig (toText databaseFile) defaultConnectionPoolConfig
   redisPool <- connect redisConfig
   timeCache <- newTimeCache simpleTimeFormat
-  (logger, rmLogger) <- newTimedFastLogger timeCache logType
+  (logger, _) <- newTimedFastLogger timeCache logType
 
   let env = Env {dbPool, redisPool, logger}
 
-  runReaderT initializeDatabase env
+  -- TODO Disable/handle auto-migration
```
sirrus233, line 58: This automatically performs database migrations if the schema changes - not good for production, but excellent for testing migrations or for the rapid dev stage.

backend/app/Main.hs
```diff
@@ -24,35 +25,38 @@ redisConfig = defaultConnectInfo
...
+  runSqlPool (runMigration migrateAll) dbPool
 
   log logger "Starting server"
   run 8081 . app $ env
-
```
sirrus233, line 54: This all cleans up when the program terminates anyway. Unnecessary, so gone.

backend/src/Database.hs
sirrus233, line 1: Note: You won't see any direct SQL statements anymore - `esqueleto` exposes a DSL for doing backend-agnostic SQL queries via the library API. Takes some getting used to, but it's all very well-typed.

backend/src/Types/Api.hs
```diff
@@ -27,10 +28,37 @@ data GameReport = GameReport
...
+      gameReportCorruption = r.corruption,
+      gameReportMordor = r.mordor,
+      gameReportInitialEyes = r.initialEyes,
+      gameReportAragornTurn = r.aragornTurn,
+      gameReportStrongholds = r.strongholds,
+      gameReportInterestRating = r.interestRating,
+      gameReportComments = r.comments
+    }
 
 data SubmitGameReportResponse = SubmitGameReportResponse
```
sirrus233, line 60: Do you need anything from this API response other than what's here?

backend/src/Types/App.hs
```diff
@@ -1,17 +1,21 @@
...
-data Env = Env {dbPool :: Pool SQL.Connection, redisPool :: Redis.Connection, logger :: TimedFastLogger}
+data Env = Env {dbPool :: SQL.ConnectionPool, redisPool :: Redis.Connection, logger :: TimedFastLogger}
 
 type AppM = ReaderT Env Handler
 
 nt :: Env -> AppM a -> Handler a
-nt s x = runReaderT x s
+nt env server = runReaderT server env
 
+runDb :: SqlPersistT IO a -> AppM a
```
sirrus233, line 16: Helper function to run SQL actions inside the AppM monad. All queries inside the action are executed in a single transaction.

backend/src/Types/Database.hs
```diff
@@ -1,117 +1,133 @@
...
+import Data.Time.Clock (getCurrentTime)
+import Database.Persist.TH (mkMigrate, mkPersist, persistLowerCase, share, sqlSettings)
+import Types.DataField (Competition, Expansion, League, Match, PlayerName, Rating, Side (..), Stronghold, Victory)
 
-data WritePlayer = WritePlayer
-  { name :: PlayerName,
-    country :: Maybe Text
-  }
-  deriving (Generic)
+share
```
sirrus233, line 11: This definition of the database probably looks like nonsense - it's something called "Template Haskell" (TH) which is basically a DSL you use to write code that writes more Haskell for you. At compile time, this custom syntax is compiled into Haskell source code, and then *that* valid Haskell is compiled. A bit confusing, but great for auto-generating boilerplate that would otherwise have to all be hand-written.

This syntax comes from `persistent` which defines all the TH rules for defining a table. When this is compiled, it auto-generates a ton of data types for the records, their fields, and associated tables.

backend/src/AppServer.hs
```diff
@@ -49,50 +46,36 @@ ratingAdjustment winner loser
...
   Failure errors -> throwError $ err422 {errBody = show errors}
-  Success (GameReport {..}) -> do
-    env <- ask
-    liftIO . withResource env.dbPool $ \conn -> withTransaction conn $ do
-      timestamp <- liftIO getCurrentTime
-      winnerId <- insertPlayerIfNotExists conn winner
-      loserId <- insertPlayerIfNotExists conn loser
-      report <- insertGameReport conn WriteProcessedGameReport {..}
+  Success (RawGameReport {..}) -> runDb $ do
+    -- TODO Logging in this stupid monad
```
chelseaSchmidt, line 53: 😂

backend/src/Types/Api.hs
```diff
@@ -27,10 +28,37 @@ data GameReport = GameReport
...
-instance FromJSON GameReport
+instance FromJSON RawGameReport
+
+instance ToJSON RawGameReport
+
+toGameReport :: UTCTime -> PlayerId -> PlayerId -> RawGameReport -> GameReport
+toGameReport timestamp winnerId loserId r =
+  GameReport
+    { gameReportTimestamp = timestamp,
+      gameReportWinnerId = winnerId,
```
chelseaSchmidt, line 39: Does the `gameReport` prefix magically map to the `game_report` table when this is run through `SQL.insert`?? So confused but this is cool haha

backend/src/Types/Api.hs
chelseaSchmidt: I don't think so, maybe a future nice-to-have would be the `pid` of any new players created? If there were to be some kind of profile page for a player, that would let us display a link to it after the submission succeeds. Not needed right now, just thinking out loud

backend/src/AppServer.hs
```diff
@@ -49,50 +46,36 @@ ratingAdjustment winner loser
...
     (smallAdjust, bigAdjust) = maybe maxThreshold snd (Map.lookupGE diff ratingThresholds)
 
-submitReportHandler :: GameReport -> AppM SubmitGameReportResponse
-submitReportHandler r = case validateReport r of
+submitReportHandler :: RawGameReport -> AppM SubmitGameReportResponse
+submitReportHandler report = case validateReport report of
   Failure errors -> throwError $ err422 {errBody = show errors}
-  Success (GameReport {..}) -> do
-    env <- ask
-    liftIO . withResource env.dbPool $ \conn -> withTransaction conn $ do
```
chelseaSchmidt, line 57: It's been fun watching the progression of `conn` being shunned more and more with each refactor 😂

backend/src/Database.hs
```diff
@@ -1,157 +1,79 @@
 module Database where
```
chelseaSchmidt, line 1: The simplicity of this file is crazy now! I think all the complexity was basically shifted off to types in `Types.Database`, which also is somehow beautifully simple looking because of Template Haskell? Huge refactor win

backend/src/Types/App.hs
chelseaSchmidt: ✨

backend/src/Types/Database.hs
chelseaSchmidt: This is serious magic 🤯

chelseaSchmidt (approved):
Wanted to get you out of feature branch hell so a lil' christmas review! All my comments are just questions to understand how everything works, good to merge! Even from a code readability standpoint alone, the refactor was worth it, this code is very pretty 🚀

backend/app/Main.hs
sirrus233: > is the migrateAll an auto-generated function from the Template Haskell in Types.Database? And migrateAll will either do migrations within an existing database if any tables contain non-matching schemas, or initialize a new database with new tables if no database exists yet?

All exactly correct!

> And does "migration" equate to generating a new table with no data - like it doesn't automagically migrate the data itself right?

I have no idea! There's a few defaults that would seem reasonable (for example, what if we add a new nullable field? Can it sorta "insert" that into the table, backfilling existing rows?), but I didn't dig much into the auto-migrate behavior. We'll have to experiment. For now - it's just super convenient that the table stays up to date without any additional effort - and without writing any SQL by hand!

backend/src/AppServer.hs
sirrus233: Hahaha, man, screw conn. All my friends hate conn.

Nah for real though, that's sort of a neat observation. I like that that's what happening, because managing a connection to the database really feels like something that Haskell should be doing for me (via some monad), and not something I should be doing by hand. Fortunately, that's now exactly the nice behavior we have!

sirrus233: I keep using "monad", "context", and "monadic context" kind of interchangeably - but I really do think "a context in which a computation happens" is the closest thing we can have to an "intuitive" understanding of such an abstract concept of a monad.

In this case, it's pretty educational to look at the type of `runDb` and see that its single parameter is of type `SqlPersistT IO a` (which is to say, an SqlPersist monad *transformer* that wraps an IO monad). The entire `do` block is the argument given to `runDb`, so the whole block happens within the `SqlPersistT` context. And what is that context? It's "a computation that has access to a database connection, where all queries on the connection happen in a single transaction". A very nice monad indeed!

backend/src/Database.hs
sirrus233: Heck yeah! The downside of TH is that it's basically a bespoke DSL every time, so you always have to learn the custom syntax of the library author. The major upside is that sometimes that syntax is terse, beautiful, and way nicer than manually writing the equivalent "real" Haskell.

The `persistent` library we're using here has a really nice documentation book. It's perhaps useful to check out this one section, which shows the TH for defining a `Person` table (very similar to what we're doing here), and then shows the actual equivalent Haskell code that is generated by the TH: https://www.yesodweb.com/book/persistent#persistent_code_generation

backend/src/Types/Api.hs
sirrus233: It's dark magic indeed, made darker by the fact that the TH obscures the real types/kinds that are going on behind the scenes. But the gist is this: the tables and the entities that populate them are compile-time related by their types. When you say `SQL.insert someGameReport`, you're not just inserting a blob of arbitrary data into a specified table... you're inserting a strongly typed record that can *only* be inserted into the table it fits in - anything else would be a type error. And because of that, type inference works everything out for us. No need to specify a table or anything. Each generated record type has one and only one table where it fits.

backend/src/Types/Api.hs
sirrus233: Great idea - easily done! I'll do that ~sometime in the future

backend/src/Types/Database.hs
sirrus233: YES

backend/src/Types/Api.hs
chelseaSchmidt: That is freakin cool. And it would be so hard for a runtime bug to get through all that. `Each generated record type has one and only one table where it fits` - perfection

backend/src/AppServer.hs
chelseaSchmidt: "A context in which a computation happens" is a helpful way to think about it for me! It's an awesome monad 👏

PR #31, opened 2024-12-24
Bugfix - Read correct player stats

sirrus233:
We have two functions for getting a player's stats - one retrieves the most recent stats from any year, the other retrieves only stats for a specified year (rolling over stats from a prior year if need-be). I used the former, but should have been the latter (since we're always submitting for the current year).

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -4,12 +4,12 @@ import Api (Api)
 import Data.IntMap.Strict qualified as Map
 import Data.Time.Clock (getCurrentTime)
 import Data.Validation (Validation (..))
-import Database (getMostRecentStats, insertGameReport, insertPlayerIfNotExists, insertRatingChange, replacePlayerStats)
+import Database (getStats, insertGameReport, insertPlayerIfNotExists, insertRatingChange, replacePlayerStats)
 import Servant (ServerError (errBody), ServerT, err422, throwError)
 import Types.Api (RawGameReport (..), SubmitGameReportResponse (..), toGameReport)
 import Types.App (AppM, runDb)
 import Types.DataField (Match (..), Rating, Side (..))
-import Types.Database (PlayerStats (..), RatingDiff (..), updatePlayerStatsLose, updatePlayerStatsWin)
+import Types.Database (PlayerStats (..), RatingDiff (..), currentYear, updatePlayerStatsLose, updatePlayerStatsWin)
 import Validation (validateReport)
 
 defaultRating :: Rating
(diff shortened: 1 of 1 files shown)


backend/src/AppServer.hs
```diff
@@ -53,11 +53,12 @@ submitReportHandler report = case validateReport report of
...
     -- TODO Reduce code duplication
     timestamp <- liftIO getCurrentTime
+    year <- currentYear
     winnerId <- insertPlayerIfNotExists winner
     loserId <- insertPlayerIfNotExists loser
     reportId <- insertGameReport $ toGameReport timestamp winnerId loserId report
-    winnerStats <- getMostRecentStats winnerId -- TODO Wrong, should be for current year specifically
-    loserStats <- getMostRecentStats loserId
+    winnerStats <- getStats winnerId year
+    loserStats <- getStats loserId year
```
chelseaSchmidt, line 61: Not a real review comment, just for learning: is the effective difference between `getMostRecentStats` and `getStats` that if no stats exist for the player in the current year yet, a db insert is done to roll the stats over from a prior year? Is this just the most convenient time and place to initialize current year stats for a player, or is it also that a player's current year stats shouldn't exist until they play their first game?

chelseaSchmidt (approved):
Softball review, 🚢!

backend/src/AppServer.hs
sirrus233: I probably could've documented this. My bad...

> if no stats exist for the player in the current year yet, a db insert is done to roll the stats over from a prior year?
Yes.
> is it also that a player's current year stats shouldn't exist until they play their first game?
Yes - in fact, a Player doesn't exist until they play their first game. Once that happens, the player and their initial stats entry are created at the same time. (See the database function for player creation)

So I had a few guiding principles here:

* A player_stats is keyed off of (playerId, year)
* A stats entry should *always* be available for a player
* I don't want to do manual or scheduled migrations of player stats every year

It's a little tricky to get all this, as made evident by the types (a very neat sentence). Database queries return a `Maybe`, as they should - so we want to dispose of that safely.

`getMostRecentStats` finds the most-recent available stats row for a given player. Since a `stats` entry is created for a player when the player is created - this should always succeed. To guarantee that, we have a failsafe where if the stats row is missing for ~some~ reason, we just create the default one (the same one we should have created at the same time as the player). Such a fallback should never happen, and should log some kind of warning/error (upcoming logging PR).

`getStats` asks for stats from a given year. This may or may not legitimately exist! If they do, then you get the stats you asked for. If they don't, then we create a new stats entry for this year, by finding the *previous* stats and rolling them forward.

As I write this - I realize there's a bug in this implementation. So - good catch by proxy! Because we only "roll forward" stats when they are queried.... suppose we have stats for 2020, 2021, 2022, and 2024. Then we query for 2023. Under the current implementation, we would "roll forward" from *2024* and not what should be expected, which is *2022*. Oops. Actually, there's a general weirdness here with the `total` rows, since those kinda stop updating once the year is over. Perhaps this schema needs some rethinking :thinking:

chelseaSchmidt: > suppose we have stats for 2020, 2021, 2022, and 2024. Then we query for 2023. Under the current implementation, we would "roll forward" from 2024

Dang solid find! Did not occur to me at all, would have been such a strange thing to debug. And thank you for explaining the rest, the solution makes total sense with those guiding principles.

PR #33, opened 2024-12-25
Support shadow stronghold reporting

sirrus233:
*NOTE:* This change won't do anything without FE support

Formerly the backend only supported reporting of captured FP strongholds in the specific case of a Shadow military victory. There was no way to report, in the event of a FPMV, which Shadow strongholds had been captured. The current Google Form report supports this, so we need to as well.

In fact, this change goes a little further than the Google Form. That form only asks for captured Free strongholds (when SPMV) **or** captured Shadow strongholds (when FPMV). In that way, it sort of loses data (what if Shadow won, but Free still caught a stronghold?). This new validation should allow for mixing of captured Sides, and still validate everything correctly. At least I think - needs some unit tests :)


chelseaSchmidt (approved):
Awesome 👏 Will add FPMV stronghold reporting on the frontend with matching values, and make both free and shadow form controls show up regardless of the winner's `side` (so `side` will just inform validation requirements). Manual FE testing will hopefully catch any unexpected issues in lieu of unit tests!

sirrus233:
Sweeeet! Actually, I was thinking - a few of these only show up / count under certain conditions:

* South Rhun / Ered Luin only count when playing with the `Cities` expansion (they are otherwise considered Towns and are not worth points, thus not relevant to us)
* Iron Hills only counts when playing with `Fate of Erebor` expansion
* Erebor is a FP stronghold normally, but when playing with `Fate of Erebor` it is a SP stronghold instead (point value remains the same either way).

All this is (I think) currently validated on the backend - but it would be cool if we enforced it in the FE form. I believe right now the stronghold chips just have a note saying e.g. "Fate of Erebor Only"

sirrus233:
Because this *is* validated in the BE, doesn't have to be MVP. Although Erebor is kind of obnoxious - considering who gets the points is dependent on expansion.

chelseaSchmidt:
I was thinking about that when I was doing the mapping from values -> display values! Like if FE can only pop those stronghold buttons up when the relevant expansion is clicked, we can also remove some of the extra display text warning that it's only applicable for the given expansion. Definitely on the to-do list even if it doesn't make the MVP cut. The `Fate of Erebor` thing will be an interesting layer haha

PR #34, opened 2024-12-26
Logging

sirrus233:
Took a bit of work to get this right, but I think I'm happy with how it turned out.

We add a few things:
* General functions and types for dealing with logging.
* Ke-jigger the application monad stack to get a Logger monad in there, allowing us to log from anywhere within the app.
* Actual log messages now that we have logging.


backend/src/Logging.hs
```diff
@@ -0,0 +1,46 @@
...
+  ( FormattedTime,
+    LogType' (..),
+    ToLogStr (..),
+    defaultBufSize,
+    newTimeCache,
+    newTimedFastLogger,
+    simpleTimeFormat,
+  )
+
+type Logger = Loc -> LogSource -> LogLevel -> LogStr -> IO ()
```
sirrus233, line 14: This is the type expected by the [MonadLogger typeclass](https://hackage.haskell.org/package/monad-logger-0.3.40/docs/Control-Monad-Logger.html#g:1) we're using. We want to leverage this typeclass because it allows us to log from anywhere within our monad transformer stack, without having to manually lift the function into the actual logging monad (very convenient).

sirrus233: In practice, we won't really use Loc or LogSource. But LogLevel is nice to use.

backend/src/Logging.hs
```diff
@@ -0,0 +1,46 @@
...
+
+type LogLevelFilter = LogSource -> LogLevel -> Bool
```
sirrus233, line 16: Another type [expected by the library](https://hackage.haskell.org/package/monad-logger-0.3.40/docs/Control-Monad-Logger.html#v:filterLogger). Nice helper function for log filtering.

backend/src/Logging.hs
```diff
@@ -0,0 +1,46 @@
...
+  | level == LevelDebug = False
+  | otherwise = True
+
+stdoutLogger :: IO Logger
+stdoutLogger = do
+  timeCache <- newTimeCache simpleTimeFormat
+  (logger, _) <- newTimedFastLogger timeCache (LogStdout defaultBufSize)
+  pure $ \_ _ level msg -> logger (\time -> formatLogMsg time level msg)
+
+log :: Logger -> LogLevel -> LogStr -> IO ()
```
sirrus233, line 42: Normally we'll use library functions to log from within our MonadLogger instance. But this is for if we need to log outside the monad for some reason (currently we use this in `main`)

backend/src/Logging.hs
```diff
@@ -0,0 +1,46 @@
...
+log logger = logger defaultLoc ""
+
+(<>:) :: (Semigroup a, IsString a, Show b) => a -> b -> a
```
sirrus233, line 45: I had an idea!

backend/src/App.hs
```diff
@@ -0,0 +1,21 @@
...
+import Database.Esqueleto.Experimental qualified as SQL
+import Database.Redis qualified as Redis
+import Logging (LogLevelFilter, Logger, filterInfo)
+import Servant (Handler)
+
+data Env = Env {dbPool :: SQL.ConnectionPool, redisPool :: Redis.Connection, logger :: Logger}
+
+type AppM = ReaderT Env (LoggingT Handler)
+
+runAppLogger :: (MonadIO m) => LogLevelFilter -> Logger -> LoggingT m a -> m a
```
sirrus233, line 14: A function like this (the critical part being the `LoggingT m a -> m a` bit) is necessary for "unwrapping" the logging part of our monad stack. Just like how `runReaderT` unwraps the Reader monad. You can see how these compose together in the `nt` function below. And you can also see how we use it to shim our logger into our (now slightly taller) SQL monad stack for runDb.

backend/src/AppServer.hs
```diff
@@ -50,10 +52,12 @@ submitReportHandler :: RawGameReport -> AppM SubmitGameReportResponse
 submitReportHandler report = case validateReport report of
   Failure errors -> throwError $ err422 {errBody = show errors}
   Success (RawGameReport {..}) -> runDb $ do
-    -- TODO Logging in this stupid monad
+    logInfoN $ "Processing game between " <> winner <> " and " <> loser <> "."
```
chelseaSchmidt, line 55: The ease/readability of this logging is totally deceptive, would never notice how it removes the lifting-into-monad effort if not for all the background you've given on monad transformer stacks 💡

backend/src/Database.hs
```diff
@@ -30,34 +32,38 @@ import Types.Database
...
       pure playerKey
 
-getStats :: (MonadIO m) => PlayerId -> Year -> SqlPersistT m PlayerStats
+getStats :: (MonadIO m, MonadLogger m) => PlayerId -> Year -> SqlPersistT m PlayerStats
 getStats pid year =
   SQL.get (PlayerStatsKey pid (fromIntegral year)) >>= \case
     Just stats -> pure stats
     Nothing -> do
-      priorStats <- getMostRecentStats pid
+      priorStats <- getMostRecentStats pid -- TODO Buggy sadness https://github.com/sirrus233/wotr-community-website/pull/31#discussion_r1897581825
```
chelseaSchmidt, line 54: 😂

backend/src/App.hs
chelseaSchmidt: ✨

chelseaSchmidt (approved):
Clean and convenient 💯 the explanatory comments were helpful!

backend/app/Main.hs
```diff
@@ -47,16 +44,17 @@ main :: IO ()
...
-  timeCache <- newTimeCache simpleTimeFormat
-  (logger, _) <- newTimedFastLogger timeCache logType
 
   let env = Env {dbPool, redisPool, logger}
 
   -- TODO Disable/handle auto-migration
-  runSqlPool (runMigration migrateAll) dbPool
+  migrations <- runSqlPool (runMigrationQuiet migrateAll) dbPool
+  unless (null migrations) (log logger LevelWarn "Database schema changed. Running migrations.")
+  mapM_ (log logger LevelInfo . toLogStr) migrations
```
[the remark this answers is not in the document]
sirrus233: Yeah, totes. The default behavior of `runMigration` is to print the executed SQL commands to stderr. What `runMigrationQuiet` does is to return them as a `[Text]` instead of printing. So yeah, basically just trying to reclaim some control over how this will display. Output on stderr will just quietly disappear when this is running on a server, so this lets us log (or not log) in our standardized format.

`mapM` is quite useful. This line does exactly what you say. It's important (for Haskell learning) to understand the general case of what's being done here: `mapM` takes as its input: 
1. a function from `Monad m => a -> m b`
2. a bunch of `a`s: `t a`

If we did a normal `map` operation, we would end up with `t (m b)`, which could very well not be what we want (*maybe* we want a bunch of monads, but probably not). In the example of the code above, I'd have `[IO ()]` which sucks. So what `mapM` gives us is the ability to run all those monadic actions (the `a -> m b` applied to all the inputs), but then collect the results and wrap them all in the monad. So:

`map`, on `t a`, given `a -> m b`, yields `t (m b)`
`mapM`, on `t a`, given `a -> m b`, yields `m (t b)`

Much better! Now, in the case of our logging stuff, we end up with `IO [()]`. Nice to have the results collected under one monad - but these results are kinda stupid. We really only wanted to run the actions for their side-effect. So that's what `mapM_` is for. The only difference between `mapM` and `mapM_` is that `mapM_` discards the results - in other words, it always returns `m ()`. It's not just IO where this is useful, We can do this *even if* our monadic actions *do* return a result. Just depends on what you want :shrug: 

One more thing to note is that `mapM` is a monadic-specific version of the more general function `traverse` (which works on Applicatives). The type signature is nearly identical, differing only in the type constraint.

`traverse :: Applicative f => (a -> f b) -> t a -> f (t b)`
`mapM :: Monad m => (a -> m b) -> t a -> f (m b)`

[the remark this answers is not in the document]
sirrus233: As a good rule of thumb:

If you end up with "a collection of results wrapped in monads", but what you wanted was "a monad wrapping a collection of results", you probably want to reach for `traverse` / `mapM`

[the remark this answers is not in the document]
chelseaSchmidt: This breakdown is helpful! I didn't think to compare `mapM` to `mapM_`, that clarifies what the underscore is distinguishing. And nice to have `traverse` and `mapM` in the back pocket

PR #35, opened 2024-12-26
/reports GET API 

sirrus233:
Does what it says! Ez pz.

I don't really know how this will scale to our actual data needs, and it can probably be made more efficient (at the very least we need an index). But it works for now!

Worth noting that the API guarantees that reports are returned in time-sorted order (newest first), which is the same order in which we'd like them displayed.


backend/src/Database.hs
sirrus233, line 1: Sorry for the mess in here - I stopped doing the SQL qualified import thing (made unnecessary by hiding the conflicting functions in the Prelude). The only actually new thing here is `getGameReports`

chelseaSchmidt: Tsk so messy! Haha jk, has to happen somewhere and it's less effort than a separate PR. I didn't know about the `hiding` keyword, that's convenient

backend/src/Database.hs
```diff
@@ -66,20 +76,33 @@ getMostRecentStats pid =
...
+getGameReports =
+  select $ do
+    (report :& winner :& loser) <-
+      from $
+        table @GameReport
+          `innerJoin` table @Player
+            `on` (\(report :& winner) -> report ^. GameReportWinnerId ==. winner ^. PlayerId)
+          `innerJoin` table @Player
+            `on` (\(report :& _ :& loser) -> report ^. GameReportLoserId ==. loser ^. PlayerId)
+    orderBy [desc (report ^. GameReportTimestamp)]
```
chelseaSchmidt, line 107: Much hoogling of new symbols here for me haha. `^.` is particularly helpful ✨

chelseaSchmidt (approved):
Big leap, exciting to see the API building up! All the previous groundwork makes this look like a breeze.

backend/src/Database.hs
sirrus233: You don't want to know how much time I spent trying to debug why this query wouldn't compile, only to eventually realize I was using `on` from `Prelude` instead of `on` from `esqueleto`

chelseaSchmidt: Noooooo haha that's awful

PR #36, opened 2024-12-27
Collect FP strongholds in form and make expansion cities conditional

chelseaSchmidt:
Follow up on discussion in https://github.com/sirrus233/wotr-community-website/pull/33!

I tested different stronghold permutations and the server validation you added works great.

One side note, the server **disallows** reporting 10 shadow VP in case of a FPRV or SPRV - should it be **allowed** because a ring-based victory is checked first?

Also I labeled the new form control `What strongholds were captured by FP?` but do you think FP should be spelled out like Shadow is? And if so do people say `the Free Peoples` or just `Free Peoples`?

### DIFF
--- a/backend/src/Types/DataField.hs
+++ b/backend/src/Types/DataField.hs
@@ -123,7 +123,7 @@ data Stronghold
   | MountGundabad
   | Angmar
   | Moria
-  | DolGoldur
+  | DolGuldur
   | Orthanc
   | Morannon
   | BaradDur
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -431,9 +431,9 @@ function getStrongholdLabel(stronghold: Stronghold): string {
         case "DolAmroth":
             return "Dol Amroth";
         case "EredLuin":
-            return "Ered Luin (Cities expansion only)";
+            return "Ered Luin";
         case "IronHills":
-            return "Iron Hills (Fate of Erebor expansion only)";
+            return "Iron Hills";
         case "MountGundabad":
             return "Mount Gundabad";
         case "DolGuldur":
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -11,9 +11,9 @@ import {
     expansions,
     leagues,
     matchTypes,
-    shadowStrongholds,
     sides,
     staticFreeStrongholds,
+    staticShadowStrongholds,
     victoryTypes,
 } from "./constants";
 import { Expansion, League, Stronghold } from "./types";
@@ -38,7 +38,8 @@ function GameReportForm() {
     ] = useFormData();
 
     const freeStrongholdOptions: Stronghold[] = staticFreeStrongholds.slice();
-    const shadowStrongholdOptions: Stronghold[] = shadowStrongholds.slice();
+    const shadowStrongholdOptions: Stronghold[] =
+        staticShadowStrongholds.slice();
 
     if (isFateOfEreborSelected) {
         shadowStrongholdOptions.push("Erebor");
@@ -49,6 +50,7 @@ function GameReportForm() {
 
     if (isCitiesSelected) {
         freeStrongholdOptions.push("EredLuin");
+        shadowStrongholdOptions.push("SouthRhun");
     }
 
     return (
@@ -445,7 +447,7 @@ function getStrongholdLabel(stronghold: Stronghold): string {
         case "FarHarad":
             return "Far Harad";
         case "SouthRhun":
-            return "South Rhun";
+            return "South Rhûn";
     }
 }
 
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -11,8 +11,8 @@ import {
     expansions,
     leagues,
     matchTypes,
-    strongholds,
     sides,
+    strongholds,
     victoryTypes,
 } from "./constants";
 import { Expansion, League, Stronghold, Side } from "./types";
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -8,20 +8,14 @@ import Sheet from "@mui/joy/Sheet";
 import Button from "@mui/joy/Button";
 import {
     competitionTypes,
-    defaultFreeStrongholds,
-    defaultShadowStrongholds,
     expansions,
     leagues,
     matchTypes,
+    strongholds,
     sides,
     victoryTypes,
 } from "./constants";
-import {
-    Expansion,
-    ExpansionWithStrongholdEffect,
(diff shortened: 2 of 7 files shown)


sirrus233:
> One side note, the server disallows reporting 10 shadow VP in case of a FPRV or SPRV - should it be allowed because a ring-based victory is checked first?

Definitely a bug. Addressed in https://github.com/sirrus233/wotr-community-website/pull/37

backend/src/Types/DataField.hs
```diff
@@ -122,7 +122,7 @@ data Stronghold
   | MountGundabad
   | Angmar
   | Moria
-  | DolGoldur
+  | DolGuldur
```
sirrus233, line 125: You can revoke my Tolkien fan-club membership now.

frontend/src/GameReportForm.tsx
```diff
@@ -24,10 +25,24 @@ import VictoryPoints from "./VictoryPoints";
...
         formData,
-        { errorOnSubmit },
+        { errorOnSubmit, isFateOfEreborSelected, isCitiesSelected },
         { handleInputChange, validateField, handleSubmit },
     ] = useFormData();
 
+    const freeStrongholdOptions: Stronghold[] = staticFreeStrongholds.slice();
+    const shadowStrongholdOptions: Stronghold[] = shadowStrongholds.slice();
+
+    if (isFateOfEreborSelected) {
```
sirrus233, line 35: The hard-coding here seems a little like a code-smell, especially since this data is also kinda contained in the mapping of `conditionalCities` later.

Do you think it would be cleaner to have some kind of functional approach here? The valid capturable territory is really just a function of what expansions are being played, after all... (and only two expansions affect it: Cities and Fate of Erebor)

sirrus233 (commented):
Top-level note:

* This looks fantastic, overall - although I did leave a maintainability thought.
* I think it was missed that `South Rhun` is only a Shadow City when playing with `Cities` expansion.
* We should be consistent in the spelling, either SP/FP or Shadow/Free

backend/src/Types/DataField.hs
chelseaSchmidt: 💀

frontend/src/GameReportForm.tsx
chelseaSchmidt: Good call! I took a stab at this and I feel alright about how `freeStrongholdOptions` and `shadowStrongholdOptions` are derived now, but wasn't sure if there was a better way to handle the`isFateOfEreborSelected` and `isCitiesSelected` variables still hanging around in `useFormData` (for the purpose of passing in `useEffect` conditions). How do you feel about it, was it a code smell there too or just in combination with the messy side effect stuff in `GameReportForm`?

frontend/src/hooks/useFormData/index.tsx
```diff
@@ -164,6 +188,13 @@ const useFormData = (): [FormData, Meta, Helpers] => {
     useControlledClearEffect(formData.didFellowshipReachMordor.value, "mordor");
     useControlledClearEffect(formData.wasAragornCrowned.value, "aragornTurn");
 
+    useStrongholdDeselectEffect(
```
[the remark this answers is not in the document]
sirrus233: I'm not sure the best way to organize this stuff, but the general design principle would be that (as much as is possible) we want a single source of truth for the information.

I wonder if the types could be simplified (just have one single type: `Settlement`, representing all cities/strongholds), and then this could be encoded somewhere in a single function? The backend has a function like this:

```hs
strongholdSide :: [Expansion] -> Stronghold -> Side
strongholdSide expansions stronghold
  | FateOfErebor `elem` expansions && stronghold == Erebor = Shadow
  | otherwise = case stronghold of
      Rivendell -> Free
      GreyHavens -> Free
      WoodlandRealm -> Free
      Lorien -> Free
      HelmsDeep -> Free
      Edoras -> Free
      MinasTirith -> Free
      DolAmroth -> Free
      Pelargir -> Free
      Shire -> Free
      Dale -> Free
      Erebor -> Free
      EredLuin -> Free
      IronHills -> Free
      MountGundabad -> Shadow
      Angmar -> Shadow
      Moria -> Shadow
      DolGoldur -> Shadow
      Orthanc -> Shadow
      Morannon -> Shadow
      BaradDur -> Shadow
      MinasMorgul -> Shadow
      Umbar -> Shadow
      FarHarad -> Shadow
      SouthRhun -> Shadow
  ```
  
  Alternatively possible useful function sigs could be:
  `canCaptureSettlement :: Side -> [Expansion] -> Settlement -> Bool`

File: frontend/src/types.ts [hunk 4]
```diff
@@ -28,6 +28,11 @@ export type OptionalField = (typeof optionalFields)[number];
 
 export type PayloadField = (typeof payloadFields)[number];
 
+export type ExpansionWithStrongholdEffect = Exclude<
```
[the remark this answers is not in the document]
chelseaSchmidt: Ended up wiping this out in favor of a better overall change, but FWIW my original reason was to keep it tied to the `Expansion` type (but you are right that the inverse would work fine and be shorter)

frontend/src/hooks/useFormData/index.tsx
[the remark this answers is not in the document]
chelseaSchmidt: Used this as a guide to update, it's a lot better now 💯 not sure if it's all the way there yet though, switched to discord for brainstormy stuff

frontend/src/hooks/useFormData/index.tsx
```diff
@@ -172,10 +196,23 @@ const useFormData = (): [FormData, Meta, Helpers] => {
     useControlledClearEffect(formData.didFellowshipReachMordor.value, "mordor");
     useControlledClearEffect(formData.wasAragornCrowned.value, "aragornTurn");
 
+    strongholds.filter(isStrongholdConditional).map((stronghold) => {
+        useStrongholdDeselectEffect(
+            stronghold,
+            isStrongholdInPlay(formData.expansions.value, stronghold)
+        );
+    });
```
chelseaSchmidt, line 204: Mentioned having issues with this bit in Discord, but then I added a filter for `isStrongholdConditional` so it will only generate the three `useEffect`s actually needed. I can live with generating them dynamically - `isStrongholdConditional` will never return a different result, so there's no danger of a runtime error, and I just was having an irrational feeling. TLDR; scratch everything I said in Discord

frontend/src/hooks/useFormData/index.tsx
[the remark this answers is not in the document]
chelseaSchmidt: Nvm - pushed one more improvement and now I'm pretty happy with it, it's ready for another look

sirrus233:
I think this looks great. Really nice work here 🥇

sirrus233 (approved):

PR #38, opened 2024-12-28
Monad refactor

sirrus233:
Was trying to write the leaderboard API and realized I wanted another refactor. So you get another one of these :grimacing:

The basic change here is that the SQL monad is now wrapped in a ExceptT monad transformer - which makes it possible for us to throw Exceptions (server errors) when something bad happens at the database level. My current thinking is that this would be reserved for cases where we expect some invariant to hold (e.g. all players have stats) but for some reason (?) it doesn't.

The main changes are in `database.hs` with the other modules being edited to support them.


backend/src/Database.hs
```diff
@@ -1,108 +1,111 @@
...
+runDb :: DBAction (LoggingT IO) a -> AppM a
+runDb dbAction = do
+  env <- ask
+  result <- liftIO . runAppLogger env.logger . runSqlPool (runExceptT dbAction) $ env.dbPool
+  case result of
+    Left err -> throwError err
+    Right a -> pure a
+
+getPlayerByName :: (MonadIO m, MonadLogger m) => PlayerName -> DBAction m (Maybe (Entity Player))
+getPlayerByName = lift . getBy . UniquePlayerName
```
sirrus233, line 57: One little annoyance is that our functions now run in `ExceptT ServerError (SqlPersistT m)`, but `esqueleto`'s library functions still run in `SqlPersistT m`. To get around this, we need to use `lift` to lift the default monadic actions from `esqueleto`'s library, into our "exception handling" version of the monad.

backend/src/Database.hs
```diff
@@ -1,108 +1,111 @@
...
-      let stats = defaultPlayerStats pid year
-      insert_ stats
-      pure stats
-  where
-    recentStats = selectOne $ do
-      stats <- from $ table @PlayerStats
-      where_ (stats ^. PlayerStatsPlayerId ==. val pid)
-      orderBy [desc (stats ^. PlayerStatsYear)]
-      pure stats
+getPlayerStats :: (MonadIO m, MonadLogger m) => PlayerId -> DBAction m (Maybe (Entity PlayerStatsTotal, Entity PlayerStatsYear))
```
sirrus233, line 73: I decided to delete a bunch of the buggy year-stat handling stuff, and just not worry about it for now (for MVP). This just reads stats for the current year, and assumes they exist. We can figure out a longer-term plan later.

backend/src/Types/Database.hs
```diff
@@ -49,85 +49,98 @@ share
     ratingAfter Rating
     deriving Show
 
-   PlayerStats
```
sirrus233, line 52: I broke `PlayerStats` into two tables, `PlayerStatsYear` for yearly stats, keyed on `(PlayerId, Year)`
and `PlayerStatsTotal` for the all-time stuff, keyed only on `PlayerId`

chelseaSchmidt (approved):
My code review was a little fast because I gotta catch up on frontend 🫠 but I read through enough for a high-level understanding - nice addition!

backend/src/AppServer.hs
```diff
@@ -48,45 +65,55 @@ ratingAdjustment winner loser
...
-    winnerStats <- getStats winnerId year
-    loserStats <- getStats loserId year
-
     reportId <- insertGameReport $ toGameReport timestamp winnerId loserId report
 
     let (winnerSide, loserSide) = (side, other side)
-    let (winnerRatingOld, loserRatingOld) = (getRating winnerSide winnerStats, getRating loserSide loserStats)
+
+    (winnerStatsTotal, winnerStatsYear) <-
+      bimapF entityVal entityVal (readOrError ("Could not find stats for " <> show winner) $ getPlayerStats winnerId)
```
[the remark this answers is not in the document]
sirrus233: Yup, nice catch ty

backend/src/Types/Database.hs
sirrus233: Certainly possible, but I think my concern is around query efficiency. The leaderboard needs to compute totals for every player every time. That said, it still might work. We should discuss 💭

sirrus233 (commented):
.

PR #39, opened 2024-12-28
Leaderboard API

sirrus233:
It does the thing! Pretty self-explanatory, I should think. And perhaps the last of the major APIs we need???

### DIFF
--- a/backend/src/Api.hs
+++ b/backend/src/Api.hs
@@ -1,11 +1,13 @@
 module Api where
 
 import Servant (Get, JSON, Post, ReqBody, (:<|>), (:>))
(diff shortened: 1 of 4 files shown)


chelseaSchmidt (approved):
Awesome!! I haven't looked at the leaderboard enough yet to know if anything's missing for frontend, but once I get into display logic I might have more productive feedback. This looks great 🎉

chelseaSchmidt:
P.S. I started the leaderboard display work, maybe LoME stats could be good to add in a follow-up? Alternatively though I could use `/reports` in combination with `/leaderboard` to fill those in on the frontend.

PR #40, opened 2024-12-29
Display success message and reset game report form after submitting

chelseaSchmidt:
Tried to do the minimum amount of work on this I could! Next will be displaying an error message after a submission.

Uncovered a sneaky bug while adding the `resetForm` effect - the number inputs were uncontrolled components. Totally missed that before!


sirrus233 (approved):
:ship: :ship: :ship: :ship: :ship: :ship:

PR #43, opened 2024-12-30
Backend Server Infra

sirrus233:
Changes in this PR:
* Relax CORS to any domain. That may or may not be what we want long term, but required for now
* Serve over HTTPS instead of HTTP
* Modify the submitReport call in the frontend to hit the real server. Will want to follow this up with a dev/prod distinction so we can develop against localhost if we want.
* EC2 instance and all associated resources for our server
* Added redirect behavior to cloudfront so the react routing happens correctly

Changes done by hand, so not shown here:
* Manually generated a Key Pair for ssh access to server
* Associated DNS record for api.wotrcommunity.net with EC2 instance IP address


chelseaSchmidt (approved):
YAY nice work! The infra is infraing 🎉

PR #45, opened 2024-12-30
Update leaderboard endpoint in frontend

chelseaSchmidt:
Took me a sad amount of time to realize this was part of the reason the server doesn't like me, that's my queue to go to bed 😂


chelseaSchmidt:
And I spelled cue like queue, omg 🤦‍♀️

PR #46, opened 2024-12-30
Deployment Scripts

sirrus233:
Adds some lovely helper scripts, now that we're ON THE INTERNET.

A deployment script for frontend and backend respectively, and a provisioning script to do all the first-time setup required if we ever swap out the EC2 instance.

There are some implicit requirements to run these:

`provision.sh`:
 * Need credentials to ssh to the host as the `ec2-user`

`deploy.sh` (Backend flavor)
* Requires cabal
* Same ssh permissions that are required to provision

`deploy.sh` (Frontend flavor)
* Requires npm
* Requires aws-cli
* Requires access to the relevant AWS resources (s3+cloudfront), stored under a "wotrcommunity" profile

### DIFF
--- a/backend/bin/provision.sh
+++ b/backend/bin/provision.sh
@@ -1,0 +1,51 @@
+#!/bin/sh
+
+set -e
+
+BIN_DIR=$(dirname "$0")
+source "${BIN_DIR}/config.sh"
+  
+ssh "$SERVER_USER@$SERVER_HOST" <<EOF
+  set -e
+  
+  mkdir -p wotr-server
+  
+  echo "Acquiring SSL certificate..."
+  sudo yum install -y certbot
+  sudo certbot certonly --standalone -d $SERVER_HOST -m $SSL_EMAIL --agree-tos --non-interactive
+  sudo systemctl enable --now certbot-renew.timer
+  
+  echo "Setting system permissions..."
+  sudo groupadd -f ssl-cert
+  sudo usermod -a -G ssl-cert ec2-user
+  sudo chgrp -R ssl-cert "/etc/letsencrypt/live"
+  sudo chgrp -R ssl-cert "/etc/letsencrypt/archive"
+  sudo chmod -R 750 "/etc/letsencrypt/live"
+  sudo chmod -R 750 "/etc/letsencrypt/archive"
+  
+  echo "Generating service file..."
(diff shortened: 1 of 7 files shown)


infrastructure/lib/infrastructure-stack.ts
```diff
@@ -11,8 +11,10 @@ export class InfrastructureStack extends cdk.Stack {
     constructor(scope: Construct, id: string, props?: cdk.StackProps) {
         super(scope, id, props);
 
+        const elasticIp = new ec2.CfnEIP(this, "ElasticIP");
```
sirrus233, line 14: Re-org'd this so we can swap out / kill the instance but keep the same IP.

backend/bin/provision.sh
```diff
@@ -0,0 +1,51 @@
...
+
+BIN_DIR=$(dirname "$0")
+source "${BIN_DIR}/config.sh"
+  
+ssh "$SERVER_USER@$SERVER_HOST" <<EOF
+  set -e
+  
+  mkdir -p wotr-server
+  
+  echo "Acquiring SSL certificate..."
```
chelseaSchmidt, line 13: I tried to figure out a script for automating a letsencrypt SSL cert renewal at one point in my life and remember just giving up in despair lol, this script is sweet and I will probably reference it in the future

frontend/bin/deploy.sh
```diff
@@ -0,0 +1,22 @@
...
+
+echo "Building the frontend..."
+npm install
+npm run build
+
+echo "Uploading files to S3..."
+aws --profile wotrcommunity s3 sync $BUILD_DIR s3://$S3_BUCKET --delete
+
+echo "Invalidating CloudFront cache..."
+aws --profile wotrcommunity cloudfront create-invalidation --distribution-id $CLOUDFRONT_DISTRIBUTION_ID --paths "/*"
```
chelseaSchmidt, line 20: So convenient ✨

chelseaSchmidt (approved):
👏

backend/bin/provision.sh
sirrus233: Lemme know if you wanna chat about any of the UNIX-y stuff!

chelseaSchmidt: 🙏

PR #48, opened 2024-12-30
Style leaderboard a little bit

chelseaSchmidt:
Also fixes the wonky horizontal scrolling, accidentally had two `overflow: auto`s in there


sirrus233 (approved):
⛵

PR #49, opened 2024-12-30
Bugfix - don't assume FPMV as default

sirrus233:
Fixes this coming up as the result when a form doesn't meet ANY victory conditions

image


chelseaSchmidt (approved):
Oooh good catch!

PR #50, opened 2024-12-30
Fix encoding issue

sirrus233:
Barad-what??

image

It's nearly 2025, so let's render utf-8 chars properly! Also added the proper accent to "Lórien", since we can display it now.


chelseaSchmidt (approved):
Haha wow, is that how it shows up in Firefox? That's what I get for not cross-browser testing 😂

sirrus233:
Yah - I think it would be important to do this even if you could somehow know that all your users were on Chrome. Without specifying an encoding, the browser has to guess - and *how* it will guess is outside of your control. So better to just tell it how the text is encoded!

PR #54, opened 2024-12-31
Display game reports

chelseaSchmidt:
Once again did this a little quickly, so there's probably stuff missing, but initial pass!

Still left to do:
- We should audit the columns against the ones we want to keep from the spreadsheet
- I want to submit more variations of game reports to make sure all the variations of summarized values get displayed as intended
- Line wrapping in the table cells might be good since a line will just extend horizontally forever now
- I just realized frontend uses `comment` and backend uses `comments` in a game report haha...will fix that. Idk if submitting a form with a comment will break it because of that, haven't tried
- Game log field isn't populated yet obvs since that's not ready yet

### DIFF
--- a/frontend/src/GameReports.tsx
+++ b/frontend/src/GameReports.tsx
@@ -1,0 +1,207 @@
+import axios from "axios";
+import React, { useEffect, useState } from "react";
+import {
+    Competition,
+    Expansion,
+    League,
+    Match,
+    ProcessedGameReport,
+    Side,
+    Stronghold,
+    Victory,
+} from "./types";
+import {
+    getExpansionLabel,
+    getLeagueLabel,
+    getStrongholdLabel,
+    strongholdPoints,
+    strongholdSide,
+} from "./utils";
+import TableView from "./TableView";
+
+export default function GameReports() {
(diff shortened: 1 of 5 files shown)


frontend/src/GameReports.tsx
```diff
@@ -0,0 +1,207 @@
...
+        setLoading(false);
+    };
+
+    const refresh = () => {
+        setError(null);
+        setLoading(true);
+        getReports();
+    };
+
+    useEffect(refresh, []);
```
chelseaSchmidt, line 47: Planning to move all this into a hook so it can be reused between `GameReports` and `Rankings`, but for now it's just copy pasted

frontend/src/types.ts
```diff
@@ -112,4 +112,30 @@ export interface LeaderboardEntry {
     yearlyWinRateShadow: number;
 }
 
+export interface ProcessedGameReport {
```
chelseaSchmidt, line 115: This could probably be a mapped type to keep it in sync with `FormData`/`ValidFormData`/`GameReportPayload`, but for now just copied it straight from the backend

sirrus233 (approved):

PR #60, opened 2025-01-02
Frontend ladder year selection

chelseaSchmidt:
Most of the red diff is just an indentation change and removal of  `entry.year === CURRENT_YEAR && ...` display logic 😬 feel free to merge or add commits to this whenever the backend breaking change is ready, if I'm not around! Otherwise will be on the lookout if this needs any coordinated changes.

Included 2023 and 2022 as disabled options just so 2024 wouldn't look too lonely up there.

image

### DIFF
--- a/frontend/src/Rankings.tsx
+++ b/frontend/src/Rankings.tsx
@@ -1,20 +1,27 @@
 import axios from "axios";
 import React, { CSSProperties, ReactNode, useEffect, useState } from "react";
+import Box from "@mui/joy/Box";
+import Button from "@mui/joy/Button";
+import ButtonGroup from "@mui/joy/ButtonGroup";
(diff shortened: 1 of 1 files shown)


sirrus233 (approved):

sirrus233:
Had to make a small tweak to get the API call to work - commit is on main.
This looks nice! We should be able to see 2023 data as well, since we have it.

For some reason, the 2024 data looks quite wrong to me, but that's probably a migration issue. I'm looking into it.

chelseaSchmidt:
Oops thanks, `params` instead of `data` - thank you for fixing!

chelseaSchmidt:
Pushed a commit to allow selecting 2023, also

PR #61, opened 2025-01-03
Limit ladder and game report table heights and make row headers sticky

chelseaSchmidt:
- Changes certain elements which don't need to responsively resize (header, the ladder year selector, and the "refresh" button above tables) to have fixed heights, in order to responsively limit the ladder and game report tables to fit the remaining vertical screen space
- Pins row headers, which was enabled by bounding the table height
  - The table `stickyHeader` prop only works on 2 levels of row headers before throwing up its hands and giving up, so I had to do a funky workaround to make all 3 row header levels of the ladder table sticky
- Adds a `styles` directory so there was a good place to keep the fixed heights required for these changes (in `sizes.ts`), and a little bit of scope creep to throw a `colors.ts` in the same place, so we can have common color palette to pull from


sirrus233:
Frontend witchcraft!! Looks lovely ❤️

sirrus233 (approved):

PR #62, opened 2025-01-03
Auto-complete player names in game report form

chelseaSchmidt:
Converts player name fields into `Autocomplete`s, and displays a warning message above each one if you type a name that isn't present ✨

The player names are kind of mocked up, they're being retrieved from the `/leaderboard` endpoint with 2024 hard-coded just to have a substantial list of names displaying.

A good way to handle new players when there's time would be a `+ add new player` option inside the autocomplete, which would bring up a popup form to create a new player, and would send a POST request to a `/player` endpoint or something like that; then the new player would be merged into the `Autocomplete` options. After that we could add validation that player names MUST exist before submitting the form.

### DIFF
--- a/frontend/src/Autocomplete.tsx
+++ b/frontend/src/Autocomplete.tsx
@@ -1,0 +1,59 @@
+import React, { ReactNode, useState } from "react";
+import Alert from "@mui/joy/Alert";
+import Box from "@mui/joy/Box";
+import MaterialAutocomplete from "@mui/joy/Autocomplete";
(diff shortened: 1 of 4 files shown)


chelseaSchmidt:
image

chelseaSchmidt:
image

chelseaSchmidt:
(drafted an issue for the third paragraph if you agree, will delete it if we don't like that plan - https://github.com/sirrus233/wotr-community-website/issues/64)

sirrus233 (approved):

PR #68, opened 2025-01-03
Migration

sirrus233:
This adds a bunch of stuff:
* Data migration ability
* Rollback ability
* A few API tweaks/enhancements
* Some code reorganization for better modularity

It probably deserves a pretty careful eye, since it touches a lot. Please consider me on-call for this one, happy to answer any questions in real-time. Sorry (per usual) about how much it grew, it kinda got away from me while I wrestled with it.


backend/.gitignore
```diff
@@ -1,2 +1,3 @@
 dist-newstyle/
 data/
+migration/
```
sirrus233, line 3: Downloaded spreadsheet data goes here.

backend/app/Migration.hs
sirrus233, line 1: Moved a lot of this code out to libraries - but the core that's still here is the important stuff. Probably more easily viewed by looking at the raw file vs. the diff (which is mostly red)

backend/src/AppServer.hs
```diff
@@ -77,53 +104,70 @@ readOrError errMsg action =
       logErrorN errMsg
       throwError err500 {errBody = show errMsg}
 
-submitReportHandler :: RawGameReport -> AppM SubmitGameReportResponse
```
sirrus233, line 80: This function has been blown to pieces. Perhaps start with its new definition, and work from there?

backend/src/Database.hs
```diff
@@ -56,69 +61,108 @@ runDb dbAction = do
...
-          `on` (\(report :& _ :& loser) -> report ^. GameReportLoserId ==. loser ^. PlayerId)
+    pure (player, (totalStats, yearStats))
+
+getInitialStats :: (MonadIO m, MonadLogger m) => DBAction m [Entity PlayerStatsInitial]
+getInitialStats = lift . select $ from $ table @PlayerStatsInitial
+
+joinedGameReports :: From (SqlExpr (Entity GameReport) :& SqlExpr (Entity Player) :& SqlExpr (Entity Player))
+joinedGameReports =
+  table @GameReport
+    `innerJoin` table @Player
```
sirrus233, line 101: Would love an opinion on if you think an INNER or LEFT join is more appropriate here

backend/src/Database.hs
sirrus233, line 1: This file is less red than it really looks - I just re-ordered some of the functions. That said, I did change a few of them too, so probably reasonable to take a look. At least the type sigs

backend/src/Validation.hs
```diff
@@ -110,10 +110,10 @@ validateVictory report
     fpmv = report.side == Free && report.victory == Military
     victory = (report.side, report.victory)
     correctVictory
-      | report.corruption >= 12 = Just (Shadow, Ring)
-      | report.mordor == Just 5 = Just (Free, Ring)
       | report.side == Free && report.victory == Concession = Just (Free, Concession)
```
sirrus233, line 113: Turns out to be a much softer world if we just consider any concession as authoritative. And I think I'm okay with that compromise. You?

backend/src/Database.hs
chelseaSchmidt: Had to look up INNER and LEFT haha, it's been a while 😬 My first instinct is LEFT because then any `null` players popping up would let us know there's a data problem + we'd never have game reports go invisibly missing from the display. But I haven't thought through how much complexity that would add since I guess there would need to be handling for the potential `null`s. Plus it's kind of theoretical if we assume the data won't get out of sync, so not sure if it's worth the effort of changing

chelseaSchmidt (approved):
Dude this was so much work, awesome job! I tried to have a careful eye, but it really looks good, I don't have any changes of substance to suggest. I'm just pleased to more or less understand what's going on 😬 

🚢 🚢 🚢

backend/src/Types/Migration.hs
```diff
@@ -0,0 +1,339 @@
...
+    corruption :: Int,
+    mordor :: Maybe Int,
+    initialEyes :: Int,
+    aragornTurn :: Maybe Int,
+    strongholds :: [Stronghold],
+    interestRating :: Int,
+    comments :: Maybe Text
+  }
+
+toParsedGameReport :: GameReportWithTrash -> ParsedGameReport
```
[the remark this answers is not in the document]
sirrus233: This was one of the things that I found awkward to get my head around, but I don't know if there's a way to simplify it - there are so many ways to represent a report:

* GameReportWithTrash - read directly from the sheet
* ParsedGameReport - read from the sheet with extraneous info removed (has winner/loser names, no IDs, and a timestamp)
* RawGameReport - type sent from the submitReport form (winner/loser names, no IDs, no timestamp)
* ProcessedGameReport - a raw game report with timestamp added and annotated with playerIDs
* GameReport - the type stored in the database

backend/app/Migration.hs
```diff
@@ -1,518 +1,80 @@
...
+migrate :: [ParsedLegacyLadderEntry] -> [ParsedGameReport] -> AppM ()
+migrate legacyEntries reports = runDb $ do
+  traverse_ insertLegacyEntry . filter (\entry -> entry.player `notElem` banList) $ legacyEntries
 
-insertPlayer :: (MonadIO m) => Text -> SqlPersistT m (Key Player)
-insertPlayer name = insert $ Player name Nothing
+  forM_ (map (\r -> (r.timestamp, toRawGameReport r)) reports) $ \(timestamp, report) -> do
+    case validateReport report of
+      Failure errs -> case errs of
+        [NoVictoryConditionMet] -> insertReport_ timestamp (report {victory = Concession})
```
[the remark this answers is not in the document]
sirrus233: We currently only have to do this for 3 reports (at time of writing):
#3762
#6124
#14042

backend/src/AppServer.hs
```diff
@@ -77,53 +104,70 @@ readOrError errMsg action =
...
+      Free -> playerStatsTotalRatingFree
+      Shadow -> playerStatsTotalRatingShadow
 
-getReportsHandler :: AppM GetReportsResponse
-getReportsHandler = runDb getGameReports <&> GetReportsResponse . map fromGameReport
+reprocessReports :: (MonadIO m, MonadLogger m) => DBAction m ()
+reprocessReports = do
+  resetStats
+  reports <- getAllGameReports
+  forM_ (reverse reports) processReport
```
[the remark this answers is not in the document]
sirrus233: Yep!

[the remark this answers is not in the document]
sirrus233: Yes, I think so. It would be good to have an index on timestamp running both asc and desc, and then we query the direction we want to save on the time spent reversing the list. 100%

This was a time-saving shortcut :grimacing: esp. because I've not done any indexing of the tables yet (since they're currently "fast enough"

PR #70, opened 2025-01-03
renamePlayer API

sirrus233:
Per discussion, this won't be admin-locked until we build auth capability.

### DIFF
--- a/backend/src/Database.hs
+++ b/backend/src/Database.hs
@@ -23,10 +23,13 @@ import Database.Esqueleto.Experimental
     runSqlPool,
     select,
     selectOne,
+    set,
     table,
+    update,
     val,
     where_,
     (&&.),
+    (=.),
     (==.),
     (?.),
     (^.),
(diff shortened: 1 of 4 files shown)


backend/src/Database.hs
```diff
@@ -153,6 +156,12 @@ repsertPlayerStats (totalStats@(PlayerStatsTotal {..}), yearStats@(PlayerStatsYe
   repsert (PlayerStatsTotalKey playerStatsTotalPlayerId) totalStats
   repsert (PlayerStatsYearKey playerStatsYearPlayerId playerStatsYearYear) yearStats
 
+updatePlayerName :: (MonadIO m, MonadLogger m) => PlayerId -> PlayerName -> DBAction m ()
+updatePlayerName pid name = lift $ do
+  update $ \player -> do
+    set player [PlayerName =. val (normalizeName name), PlayerDisplayName =. val name]
```
chelseaSchmidt, line 162: Because of the uniqueness guards you have, will this throw an error if the name is already taken? I can make sure the frontend handles that specific error with a nice message

chelseaSchmidt (approved):

backend/src/Database.hs
sirrus233: Yes, but I don't actually know what error it will throw! Should figure that out, and maybe do some explicit error handling. Thanks, I'll take a look

chelseaSchmidt:
Your PR gifs are so on point 😂

PR #71, opened 2025-01-03
Modify report API

sirrus233:
This one's a little less trivial, but should do all the needful. It needs to be tested still - that's a little annoying without frontend support, so YOLO for now!

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -12,6 +12,7 @@ import Database
     getAllGameReports,
     getAllStats,
     getGameReports,
+    getPlayerByName,
     getPlayerStats,
     insertGameReport,
     insertPlayerIfNotExists,
@@ -20,13 +21,14 @@ import Database
     runDb,
     updatePlayerName,
   )
-import Database.Esqueleto.Experimental (Entity (..))
+import Database.Esqueleto.Experimental (Entity (..), PersistStoreRead (..), PersistStoreWrite (..))
 import Logging ((<>:))
-import Servant (NoContent (..), ServerError (errBody), ServerT, err422, err500, throwError, type (:<|>) (..))
+import Servant (NoContent (..), ServerError (errBody), ServerT, err404, err422, throwError, type (:<|>) (..))
 import Types.Api
   ( GetLeaderboardResponse (GetLeaderboardResponse),
     GetReportsResponse (GetReportsResponse),
     LeaderboardEntry (..),
+    ModifyReportRequest (..),
     ProcessedGameReport,
     RawGameReport (..),
     RenamePlayerRequest (..),
@@ -50,6 +52,7 @@ import Types.Database
     updatedPlayerStatsWin,
   )
 import Validation (validateReport)
+import Prelude hiding (get, on)
 
 defaultRating :: Rating
 defaultRating = 500
@@ -104,7 +107,7 @@ readOrError errMsg action =
     Just a -> pure a
     Nothing -> do
       logErrorN errMsg
-      throwError err500 {errBody = show errMsg}
+      throwError err404 {errBody = show errMsg}
 
 insertReport :: (MonadIO m, MonadLogger m) => UTCTime -> RawGameReport -> DBAction m ReportInsertion
 insertReport timestamp rawReport = do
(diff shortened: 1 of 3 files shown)


backend/src/AppServer.hs
```diff
@@ -174,9 +177,33 @@ getLeaderboardHandler year =
...
 
+adminModifyReportHandler :: ModifyReportRequest -> AppM NoContent
+adminModifyReportHandler ModifyReportRequest {rid, report} = case validateReport report of
+  Failure errors -> throwError $ err422 {errBody = show errors}
+  Success _ -> runDb $ do
+    oldReport <- readOrError ("Cannot find report " <>: rid) $ lift . get $ rid
+    Entity newWinnerId _ <- readOrError ("Cannot find player " <>: report.winner) $ getPlayerByName report.winner
+    Entity newLoserId _ <- readOrError ("Cannot find player " <>: report.loser) $ getPlayerByName report.loser
+
+    let newReport = toGameReport oldReport.gameReportTimestamp newWinnerId newLoserId report
```
sirrus233, line 188: Note: the timestamp is unmodifiable, currently. We can change that in the future, if we want.

backend/src/AppServer.hs
```diff
@@ -174,9 +177,33 @@ getLeaderboardHandler year =
 adminRenamePlayerHandler :: RenamePlayerRequest -> AppM NoContent
 adminRenamePlayerHandler RenamePlayerRequest {pid, newName} = runDb $ updatePlayerName pid newName >> pure NoContent
 
+adminModifyReportHandler :: ModifyReportRequest -> AppM NoContent
+adminModifyReportHandler ModifyReportRequest {rid, report} = case validateReport report of
+  Failure errors -> throwError $ err422 {errBody = show errors}
+  Success _ -> runDb $ do
+    oldReport <- readOrError ("Cannot find report " <>: rid) $ lift . get $ rid
+    Entity newWinnerId _ <- readOrError ("Cannot find player " <>: report.winner) $ getPlayerByName report.winner
+    Entity newLoserId _ <- readOrError ("Cannot find player " <>: report.loser) $ getPlayerByName report.loser
```
chelseaSchmidt, line 186: Idk how to handle the specific case of someone entering an **existing player** in the original report, then an admin wanting to reassociate the report with a **new player** that doesn't exist yet. Is that a realistic thing that ever happens?

The combination of APIs you just added would already let an admin work around this though which is sweet. Submit a dummy report with the new player, reassociate the target report to the new player, then delete the dummy report

backend/src/AppServer.hs
```diff
@@ -174,9 +177,33 @@ getLeaderboardHandler year =
...
+
+    pure NoContent
+  where
+    mustReprocess :: GameReport -> GameReport -> Bool
+    mustReprocess old new
+      | old.gameReportWinnerId /= new.gameReportWinnerId = True
+      | old.gameReportLoserId /= new.gameReportLoserId = True
+      | old.gameReportSide /= new.gameReportSide = True
+      | old.gameReportCompetition /= new.gameReportCompetition = True
+      | otherwise = False
```
chelseaSchmidt, line 201: Beautiful 👏

backend/src/AppServer.hs
```diff
@@ -174,9 +177,33 @@ getLeaderboardHandler year =
...
+    when (mustReprocess oldReport newReport) reprocessReports
+
+    pure NoContent
+  where
+    mustReprocess :: GameReport -> GameReport -> Bool
+    mustReprocess old new
+      | old.gameReportWinnerId /= new.gameReportWinnerId = True
+      | old.gameReportLoserId /= new.gameReportLoserId = True
+      | old.gameReportSide /= new.gameReportSide = True
+      | old.gameReportCompetition /= new.gameReportCompetition = True
```
chelseaSchmidt, line 200: If `report.competition` changes from [League] -> [League, Tournament], or [Tournament] -> [League], or [League] -> [Tournament], would there be an unnecessary reprocess? `report.match` could be used to avoid that maybs. But also good to err on the side of reprocessing

backend/src/AppServer.hs
chelseaSchmidt: 🕥

chelseaSchmidt (approved):
Very nice! You're on a roll with these 💃

backend/src/AppServer.hs
sirrus233: That is IMO an incredibly unlikely event - at the very least I've never seen it happen. I think not worth addressing yet (esp. because of your identified workaround). That said, this is a really nice catch of a weird corner case, and it's good to have it back-of-mind.

backend/src/AppServer.hs
sirrus233: Well.... I definitely *meant* to put match here instead of competition. :face_exhaling:

PR #72, opened 2025-01-03
Delete report API

sirrus233:
Tested this one, works a treat! Measured the time it takes to reprocessing reports as ~roughly 10 seconds. We can probably get that down by a lot, but I think it's acceptable for now.

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -9,6 +9,7 @@ import Data.Time.Clock (getCurrentTime)
 import Data.Validation (Validation (..))
 import Database
   ( DBAction,
+    deleteGameReport,
     getAllGameReports,
     getAllStats,
     getGameReports,
@@ -40,6 +41,7 @@ import Types.Api
 import Types.DataField (Match (..), Rating, Side (..), Year)
 import Types.Database
   ( GameReport (..),
+    GameReportId,
     MaybePlayerStats,
     Player (..),
     PlayerId,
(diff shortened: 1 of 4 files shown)


backend/src/AppServer.hs
```diff
@@ -200,10 +202,17 @@ adminModifyReportHandler ModifyReportRequest {rid, report} = case validateReport
       | old.gameReportCompetition /= new.gameReportCompetition = True
       | otherwise = False
 
+adminDeleteReportHandler :: DeleteReportRequest -> AppM NoContent
+adminDeleteReportHandler DeleteReportRequest {rid} = runDb $ do
+  deleteGameReport rid
+  reprocessReports
```
chelseaSchmidt, line 208: Much work went into making it possible for this to be so simple 💯

backend/src/Database.hs
```diff
@@ -162,6 +162,12 @@ updatePlayerName pid name = lift $ do
     set player [PlayerName =. val (normalizeName name), PlayerDisplayName =. val name]
     where_ (player ^. PlayerId ==. val pid)
 
+deleteGameReport :: (MonadIO m, MonadLogger m) => Key GameReport -> DBAction m ()
+deleteGameReport rid = lift $ do
+  delete $ do
+    report <- from $ table @GameReport
+    where_ (report ^. GameReportId ==. val rid)
```
chelseaSchmidt, line 169: Is this a no-op if no such `rid` is found, or will it throw an error?

chelseaSchmidt (approved):

backend/src/Database.hs
sirrus233: Currently a no-op. Should we respond with an error? That seems like a good idea

chelseaSchmidt: No-op seems fine, could just have the response include a 1 or 0 for number of reports deleted in that case? The frontend could interpret a 0 as a failure 👌

PR #75, opened 2025-01-04
Update active status when reprocessing reports

sirrus233:
A little jank, perhaps, but I think a reasonable middle ground to start with. This adds `isActive` to the leaderboard response, and provides a database function for updating the active status across the whole table. This only runs when reports are reprocessed.

I figure that's okay for MVP, since reports are reprocessed during migration. But eventually we may want to play around with scheduling this, or doing an update check for the involved players when a report is submitted.


chelseaSchmidt (approved):
🚀

Issue #67, opened 2025-01-03
Move `GET` endpoint requests to root app level to reduce network calls


chelseaSchmidt:
Would also allow sharing data between components, like `/leaderboard` results

PR #79, opened 2025-01-05
Refactor useFormData hook to accept generic form data

chelseaSchmidt:
This is a beast sorry, it's mostly just shifting code around! I managed to split this up from the player-edit buttons change, so hopefully that makes it easier to review. Edit buttons will be out in a sec, on top of this branch.

This changes `useFormData` to be usable by any of our forms so the state management can be consistent.

  - The hook was providing `GameReportForm` with a `/leaderboard` result in order to populate the `loser` and `winner` field autocompletion, so the `/leaderboard` request is now moved into `App`, which passes the result down to both `GameReportForm` and `Rankings` (which incidentally makes progress toward [#67](https://github.com/sirrus233/wotr-community-website/issues/67))

  - The `controlledClearEffect` and other game-form-specific effect code is now in a separate hook called `useGameReportFormEffects`

  - `initialFormData` is renamed to `initialGameReportFormData` and moved out of the `hooks` directory

  - The hook now accepts as arguments: `initialFormData`, a `submit` handler, an `optionalFields` list, and a `toErrorMessage` util. These are all dependent on which form is using the hook

  - `ConstrainedFormData` was the major interesting type that I had to work out to make this possible without losing any type safety

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,6 +1,7 @@
+import axios from "axios";
 import { CssVarsProvider } from "@mui/joy/styles";
 import CssBaseline from "@mui/joy/CssBaseline";
-import React, { ReactNode } from "react";
+import React, { ReactNode, useEffect, useState } from "react";
 import GameReportForm from "./GameReportForm";
 import GameReports from "./GameReports";
 import Rankings from "./Rankings";
@@ -16,8 +17,32 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import Typography from "@mui/joy/Typography";
 import { BrowserRouter, Routes, Route, Link } from "react-router-dom";
 import { HEADER_HEIGHT_PX, HEADER_MARGIN_PX } from "./styles/sizes";
+import { Year, LeaderboardEntry } from "./types";
+import { DEFAULT_YEAR } from "./constants";
 
 export default function App() {
+    const [leaderboard, setLeaderboard] = useState<LeaderboardEntry[]>([]);
+    const [leaderboardYear, setLeaderboardYear] = useState<Year>(DEFAULT_YEAR);
+    const [loadingLeaderboard, setLoadingLeaderboard] = useState(false);
+
+    const getLeaderboard = () => {
+        setLoadingLeaderboard(true);
+
+        axios
+            .get("https://api.waroftheringcommunity.net:8080/leaderboard", {
+                params: { year: leaderboardYear },
+            })
+            .then((response) => {
+                setLeaderboard(response.data.entries as LeaderboardEntry[]);
+            })
+            .catch(console.error)
+            .finally(() => {
+                setLoadingLeaderboard(false);
+            });
+    };
+
+    useEffect(getLeaderboard, [leaderboardYear]);
+
     return (
         <CssVarsProvider>
             <CssBaseline />
@@ -52,9 +77,28 @@ export default function App() {
                 </header>
                 <Routes>
                     <Route path="/" element={<Home />} />
-                    <Route path="/game-report" element={<GameReportForm />} />
+                    <Route
+                        path="/game-report"
+                        element={
+                            <GameReportForm
+                                leaderboard={leaderboard}
+                                loadingLeaderboard={loadingLeaderboard}
+                            />
+                        }
+                    />
                     <Route path="/game-reports" element={<GameReports />} />
-                    <Route path="/rankings" element={<Rankings />} />
+                    <Route
+                        path="/rankings"
+                        element={
+                            <Rankings
+                                leaderboard={leaderboard}
+                                year={leaderboardYear}
+                                loading={loadingLeaderboard}
+                                getLeaderboard={getLeaderboard}
+                                setYear={setLeaderboardYear}
+                            />
+                        }
+                    />
                 </Routes>
             </BrowserRouter>
         </CssVarsProvider>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -17,12 +17,13 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
 import Typography from "@mui/joy/Typography";
 import { BrowserRouter, Routes, Route, Link } from "react-router-dom";
 import { HEADER_HEIGHT_PX, HEADER_MARGIN_PX } from "./styles/sizes";
-import { Year, LeaderboardEntry } from "./types";
-import { DEFAULT_YEAR } from "./constants";
+import { LeaderboardEntry } from "./types";
 
 export default function App() {
     const [leaderboard, setLeaderboard] = useState<LeaderboardEntry[]>([]);
-    const [leaderboardYear, setLeaderboardYear] = useState<Year>(DEFAULT_YEAR);
+    const [leaderboardYear, setLeaderboardYear] = useState(
+        new Date().getFullYear()
+    );
     const [loadingLeaderboard, setLoadingLeaderboard] = useState(false);
 
     const getLeaderboard = () => {
--- a/frontend/src/GameReportForm.tsx
+++ b/frontend/src/GameReportForm.tsx
@@ -1,4 +1,5 @@
 import "@fontsource/inter";
+import axios from "axios";
 import React from "react";
 import CircularProgress from "@mui/joy/CircularProgress";
 import Modal from "@mui/joy/Modal";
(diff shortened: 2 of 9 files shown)


frontend/src/App.tsx
```diff
@@ -16,8 +17,32 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
...
 
 export default function App() {
+    const [leaderboard, setLeaderboard] = useState<LeaderboardEntry[]>([]);
+    const [leaderboardYear, setLeaderboardYear] = useState<Year>(DEFAULT_YEAR);
+    const [loadingLeaderboard, setLoadingLeaderboard] = useState(false);
+
+    const getLeaderboard = () => {
+        setLoadingLeaderboard(true);
+
+        axios
```
sirrus233, line 31: Not now, but note for the future - we should configure these calls to read an environment variable (maybe `DEVELOPMENT` or something like that?) and hit `localhost` if it's set. Right now we can only really test against prod :sweat_smile:

frontend/src/App.tsx
```diff
@@ -16,8 +17,32 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
...
+            .get("https://api.waroftheringcommunity.net:8080/leaderboard", {
+                params: { year: leaderboardYear },
+            })
+            .then((response) => {
+                setLeaderboard(response.data.entries as LeaderboardEntry[]);
```
sirrus233, line 36: We should absolutely look into https://docs.servant.dev/en/stable/tutorial/Javascript.html particularly https://docs.servant.dev/en/stable/tutorial/Javascript.html#axios-support

Would be great to not risk breaking this every/any time the API changes.

frontend/src/App.tsx
```diff
@@ -16,8 +17,32 @@ import OpenInNewIcon from "@mui/icons-material/OpenInNew";
...
 import { HEADER_HEIGHT_PX, HEADER_MARGIN_PX } from "./styles/sizes";
+import { Year, LeaderboardEntry } from "./types";
+import { DEFAULT_YEAR } from "./constants";
 
 export default function App() {
+    const [leaderboard, setLeaderboard] = useState<LeaderboardEntry[]>([]);
+    const [leaderboardYear, setLeaderboardYear] = useState<Year>(DEFAULT_YEAR);
+    const [loadingLeaderboard, setLoadingLeaderboard] = useState(false);
+
+    const getLeaderboard = () => {
```
sirrus233, line 28: Is there a reason we define this here and pass it to the component, but don't do the same for the list of game reports?

frontend/src/GameReportForm.tsx
```diff
@@ -435,3 +468,110 @@ function GameReportForm() {
...
+        if (parsedResult.length) {
+            const errorMessage = parsedResult
+                .map(validationErrorToMessage)
+                .join(", ");
+            return (
+                errorMessage.slice(0, 1).toUpperCase() + errorMessage.slice(1)
+            );
+        }
+    }
+    return "Something went wrong.";
```
sirrus233, line 517: Do we want to put contact information in these error messages??

frontend/src/hooks/useGameReportFormEffects.tsx
sirrus233, line 1: These controlled clear effects are lovely! Fantastic stuff.

sirrus233 (commented):
Looks great, dude. Awesome work :heart: :rocket:

frontend/src/GameReportForm.tsx
chelseaSchmidt: Haha totes, let's stick an email in there, what's good?

frontend/src/utils.tsx
```diff
@@ -140,3 +140,58 @@ export function strongholdPoints(stronghold: Stronghold): 1 | 2 {
...
+        typeof error.status === "number" &&
+        "response" in error &&
+        typeof error.response === "object" &&
+        error.response !== null &&
+        "data" in error.response &&
+        typeof error.response.data === "string"
+    );
+}
+
+export function objectKeys<T extends object>(obj: T): Array<keyof T> {
```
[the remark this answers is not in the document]
chelseaSchmidt: Wanted to do it the functional way 🥲 would it be better to do a for...in to avoid typecasting do you think?

frontend/src/utils.tsx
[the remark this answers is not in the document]
chelseaSchmidt: Haha I totally get it, it really was a junk drawer moment, I just had to have them someplace common and didn't know what to call it. Maybe could break this out into some classes with static functions when there's some cleanup time? Like a `GameReportLabeler` or `StrongholdMetadata` class or things like that 🤔

frontend/src/constants.ts
```diff
@@ -106,3 +106,5 @@ export enum ErrorMessage {
 }
 
 export const INFINITE = 100;
+export const DEFAULT_YEAR = 2024;
```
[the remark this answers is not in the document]
chelseaSchmidt: Totes true! I guess it's time for me to stop being lazy about the year, will slide a fix in for this before I merge

frontend/src/App.tsx
chelseaSchmidt: We should also for game reports! Wasn't required by this PR and I didn't want to blow it up even more, but that's on the agenda

frontend/src/App.tsx
chelseaSchmidt: Agree! Made an issue for it: https://github.com/sirrus233/wotr-community-website/issues/82

frontend/src/Rankings.tsx
```diff
@@ -13,9 +13,13 @@ import {
...
 import TableView from "./TableView";
 
-const DEFAULT_YEAR = 2024;
-const AVAILABLE_YEARS = [DEFAULT_YEAR, 2023];
-const COMING_SOON_YEARS = [2022];
+interface Props {
+    leaderboard: LeaderboardEntry[];
+    year: Year;
+    loading: boolean;
+    getLeaderboard: () => void;
```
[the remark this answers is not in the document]
chelseaSchmidt: Yeah! Or specifically, it lets us re-**mount** the component without requerying the data, so now flipping the page back and forth doesn't trigger a refresh

frontend/src/App.tsx
chelseaSchmidt: This would be next-level 💯 Added an issue for it so I don't forget! As a stopgap I should really add types for server responses to make it more noticeable when frontend and backend types get out of sync. Having to dig through the code for axios calls is not great

frontend/src/utils.tsx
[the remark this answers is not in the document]
sirrus233: I'm just wondering if this actually gives you anything that `for..in` doesn't. I would *assume* (but do not know) that `for..in` iterating over some object `T` would consider the iterating variable of type `keyof T` -- in which case, it doesn't seem worth it to introduce more syntax to get no additional type safety.

If, on the other hand, this is doing something extra -- then by all means go for it! But I'd be curious what that is.

frontend/src/App.tsx
sirrus233: Agreed - but maybe it's not worth doing that until we are auto-generating the calls anyway? This could be something we investigate together.

frontend/src/GameReportForm.tsx
sirrus233: My personal email works, `sirrus233@gmail.com`

sirrus233: No, better idea: Let's point them to the #ladder-support channel on the discord!

frontend/src/hooks/useGameReportFormEffects.tsx
sirrus233: :joy:

frontend/src/App.tsx
chelseaSchmidt: Sweet won't memorialize it in my sticky note then, sounds like a fun project

frontend/src/constants.ts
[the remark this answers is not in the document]
chelseaSchmidt: I added a change for this - but it doesn't capture anything to do with time zones. Do you think it's fine to start with? https://github.com/sirrus233/wotr-community-website/pull/79/commits/8ccb512539ccc588c9ae5d43ad10e9c5850d0193

frontend/src/utils.tsx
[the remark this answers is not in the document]
chelseaSchmidt: Didn't have any clue, did a little experiment:
```
const obj: { a: number; b: number } = { a: 1, b: 1 };

for (const x in obj) {
    console.log(obj[x]);
}
```
And I get a MFin type error on `obj[x]`! Typescript sucks

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ a: number; b: number; }'.
  No index signature with a parameter of type 'string' was found on type '{ a: number; b: number; }'.
```

frontend/src/GameReportForm.tsx
chelseaSchmidt: Perfect 💯 grabbed link from you but gonna do this in a follow-up PR so I can rebase the other branch easier

frontend/src/constants.ts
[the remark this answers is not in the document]
chelseaSchmidt: Gonna merge so I can fix my conflicts on other branches, but time zone consideration might be a good issue for me to add, since I'm all about the issues today

PR #80, opened 2025-01-05
Frontend controls for editing and remapping of players

chelseaSchmidt:
- On leaderboard page, adds an edit pencil/`PlayerEditForm` and merge icon/`PlayerRemapForm`, for editing a player's info and remapping a player, respectively

- Expands the `Autocomplete` to allow more complex options (I went with `(string | PlayerOption)[]` for right now, but really it should be more generic, like `(string | Option)[]` and then map `playerOptions` to the generic interface. That can always be improved upon when we need an autocomplete somewhere else, I just got tired

- For the `PlayerRemapForm`, when you click submit, it's a fake-out the first time, and a "danger" alert gets displayed. The alert tells you to click submit again if you're sure

- The input in both forms get the standard "Required" validation, and one layer on top of that, which disallows the input to match the edited player's name

### DIFF
--- a/frontend/src/PlayerEditForm.tsx
+++ b/frontend/src/PlayerEditForm.tsx
@@ -1,0 +1,146 @@
+import axios from "axios";
+import React, { useEffect } from "react";
+import Box from "@mui/joy/Box";
+import Button from "@mui/joy/Button";
+import CircularProgress from "@mui/joy/CircularProgress";
+import FormControl from "@mui/joy/FormControl";
+import FormHelperText from "@mui/joy/FormHelperText";
+import FormLabel from "@mui/joy/FormLabel";
+import Sheet from "@mui/joy/Sheet";
+import Typography from "@mui/joy/Typography";
+import {
+    PlayerEditFormData,
+    ServerErrorBody,
+    ValidPlayerEditFormData,
+} from "./types";
+import TextInput from "./TextInput";
+import useFormData from "./hooks/useFormData";
+
+interface Props {
+    pid: number;
+    name: string;
+    refresh: () => void;
+}
+
+export default function PlayerEditForm({ pid, name, refresh }: Props) {
+    const initialFormData: PlayerEditFormData = {
+        pid: {
+            value: pid,
+            error: null,
+            validate: () => null,
+        },
+        newName: {
+            value: null,
+            error: null,
+            validate: function _() {
+                return this.value?.toLowerCase().trim() ===
+                    name.toLowerCase().trim()
+                    ? "Cannot rename a player to the same name"
+                    : null;
+            },
+        },
+    };
+
+    const [
+        formData,
+        { errorOnSubmit, successMessage, loading },
+        { handleInputChange, validateField, handleSubmit, setSuccessMessage },
+    ] = useFormData<PlayerEditFormData, ValidPlayerEditFormData>({
+        initialFormData,
+        optionalFields: [],
+        submit,
+        toErrorMessage,
+    });
+
+    useEffect(
+        function refreshOnSubmit() {
+            if (successMessage) {
+                refresh();
+            }
+        },
+        [successMessage]
+    );
+
+    return (
+        <Sheet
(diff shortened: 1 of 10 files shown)


frontend/src/PlayerRemapForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+                    <Button
+                        onClick={() => {
+                            if (warningAlert || formData.toPlayer.error) {
+                                handleSubmit();
+                            } else {
+                                setWarningAlert(
+                                    `Danger: ${name} will cease to exist. ${formData.toPlayer.value?.label} will absorb all of ${name}'s history. ${name}'s history cannot be recovered. If you're sure, press "Submit" again to continue.`
+                                );
+                            }
+                        }}
```
chelseaSchmidt, line 130: This is the fake-out alert step. There's a small UI bug I didn't have the energy to work out. It happens if you follow these steps:

1. Submit the form, read "danger" alert, submit again. Receive an error from the server. The error is displayed (expected)

2. Submit again. "Danger" alert is displayed (expected). But, the old submission error from the server is not cleared

This is because errors are cleared by the `handleSubmit` provided by `useFormData`, which isn't called until after the alert step. I think the solution is pass the alert message into `useFormData` so `useFormData` can handle it in the proper order. Unless you think it's urgent, will make an issue and handle it in a follow-up, maybe post-MVP.

chelseaSchmidt:
One more callout - the thing with input fields not rerunning validation until you `blur` is especially noticeable with these new forms I think. I had a working solution for it before I did the `useFormData` refactor, so I just need to figure out how to do it in the post-refactor world. Another planned follow-up

frontend/src/hooks/useFormData/index.tsx
```diff
@@ -28,7 +28,7 @@ interface Args<F, V> {
     initialFormData: ConstrainedFormData<F>;
     optionalFields: string[];
     missingFieldErrorMessage?: ErrorMessage;
-    submit: (validatedFormData: ConstrainedFormData<V>) => Promise<any>; // FIXME
+    submit: (validatedFormData: ConstrainedFormData<V>) => Promise<any>;
```
chelseaSchmidt, line 31: I should add a type for server responses, but `any` is being inferred on `main` as well, so it's not worse than before (just more obvious now that there's a type annotation). Will make another little issue for this

frontend/src/PlayerEditForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+            error: null,
+            validate: () => null,
+        },
+        newName: {
+            value: null,
+            error: null,
+            validate: function _() {
+                return this.value?.toLowerCase().trim() ===
+                    name.toLowerCase().trim()
+                    ? "Cannot rename a player to the same name"
```
sirrus233, line 38: How are we not using "Fool of a Took!!" more in these strings??

frontend/src/PlayerRemapForm.tsx
sirrus233: I don't think it's urgent! This looks good to me.

frontend/src/hooks/useFormData/index.tsx
sirrus233: Remember we can change/restructure the server responses at any point if it makes it easier to work with on the frontend. We're in full control here.

sirrus233 (approved):
Fantastic work - I can really see the refactor work shining through here. That was a great call.

frontend/src/PlayerEditForm.tsx
chelseaSchmidt: That's such a good idea 😂 we should have an issue for adding LOTR references everywhere

frontend/src/hooks/useFormData/index.tsx
chelseaSchmidt: Yes and also given Servant plan on other branch, better to leave this alone for now! Covered in all respects

PR #81, opened 2025-01-05
Add filter buttons for player active status to leaderboard

chelseaSchmidt:
Adds buttons above leaderboard table for filtering on active/inactive player status. Your API works great!

Includes a `Filters` component (a ripoff of `MultiOptionInput` for the most part) that we can reuse for more table filtering in the future.

Toggling between the filters is unbearably slow, MUI tables rendering performance strikes again 😬


sirrus233 (approved):
LGTM! I want to see the supposed slowness in action - presumably there's something we can do about that.

PR #84, opened 2025-01-05
Support submission of game logs

sirrus233:
Heck yeah let's gooo!

For frontend considerations, all that really matters is the API changes to `submitReport` and the associated request/response types.

As far as backend goes:
AWS actions operate with the help of an `Env` struct which I've now created during startup and then store in the Env of AppM. Once you're running, it's all just IO so pretty simple.

I've created a public readable bucket in our infrastructure - and the last thing I need to do is deploy putObject credentials to our host - the AWS library we're using should find those in the environment, but we'll need to test that actually happens.

### DIFF
--- a/backend/.vscode/settings.json
+++ b/backend/.vscode/settings.json
@@ -1,6 +1,7 @@
 {
     "cSpell.words": [
         "Aeson",
+        "Amazonka",
         "Amroth",
         "Angmar",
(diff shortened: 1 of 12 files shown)


chelseaSchmidt (approved):
Looks awesome, no notes, and works like a charm!

PR #124, opened 2025-01-07
Admin form layout component

chelseaSchmidt:
Just moves all the repeated form layout structure into one spot. The copy/pasting was bothering me 🫠


sirrus233 (approved):
🚲🎨

PR #136, opened 2025-01-07
Support pagination for getReports API

sirrus233:
Addresses the backend part of #112 

I can change this to a simpler query param of `page` that accepts just a number, but I figured the increased flexibility here might be nice, in case frontend would like to control how many results per page. LMK if you want it changed though!

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -198,8 +198,14 @@ submitReportHandler (SubmitReportRequest rawReport logFileData) = do
       mapM_ (putS3Object awsEnv key . fdPayload) logFileData
       pure SubmitGameReportResponse {report = processedReport, winnerRating, loserRating}
 
-getReportsHandler :: AppM GetReportsResponse
-getReportsHandler = runDb $ getGameReports 500 0 <&> GetReportsResponse . map fromGameReport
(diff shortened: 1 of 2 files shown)


backend/src/AppServer.hs
```diff
@@ -198,8 +198,14 @@ submitReportHandler (SubmitReportRequest rawReport logFileData) = do
       mapM_ (putS3Object awsEnv key . fdPayload) logFileData
       pure SubmitGameReportResponse {report = processedReport, winnerRating, loserRating}
 
-getReportsHandler :: AppM GetReportsResponse
-getReportsHandler = runDb $ getGameReports 500 0 <&> GetReportsResponse . map fromGameReport
+getReportsHandler :: Maybe Int64 -> Maybe Int64 -> AppM GetReportsResponse
+getReportsHandler limit offset = GetReportsResponse . map fromGameReport <$> runDb (getGameReports limit' offset')
+  where
+    maxLimit = 500
+    (limit', offset') = case (limit, offset) of
```
chelseaSchmidt, line 205: `limit`/`offset` works great, no specific desire for sending a page number! This turned out to be really simple for backend, nice 🎉

chelseaSchmidt (approved):

PR #140, opened 2025-01-08
Support modificatin of report timestamp

### LINKED ISSUE #116: Allow changing the timestamp of an existing report

sirrus233:
Resolves #116 

Reports being reported in the wrong order has been known to happen, so we should support modification of report timestamp along with everything else. I didn't do this initially because timestamp isn't part of a Raw Report (which was basically our input type), but it turned out to be simple enough to add.

Not sure how the UI should support this. If it's hard, we could punt there if need-be.

One note on the frontend side, [per the documentation](https://hackage.haskell.org/package/text-iso8601-0.1.1/docs/Data-Time-FromText.html#v:parseUTCTime) the valid string-serialized forms of a UTCTime are:

> YYYY-MM-DD HH:MMZ
YYYY-MM-DD HH:MM:SSZ
YYYY-MM-DD HH:MM:SS.SSSZ

> The first space may instead be a T, and the second space is optional. The Z represents UTC. The Z may be replaced with a time zone offset of the form +0000 or -08:00, where the first two digits are hours, the : is optional and the second two digits (also optional) are minutes.

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -255,6 +255,7 @@ adminModifyReportHandler ModifyReportRequest {rid, timestamp, report} = case val
   where
     mustReprocess :: GameReport -> GameReport -> Bool
     mustReprocess old new
+      | old.gameReportTimestamp /= new.gameReportTimestamp = True
       | old.gameReportWinnerId /= new.gameReportWinnerId = True
       | old.gameReportLoserId /= new.gameReportLoserId = True
       | old.gameReportSide /= new.gameReportSide = True
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -238,14 +238,15 @@ adminRemapPlayerHandler RemapPlayerRequest {fromPid, toPid} = runDb $ do
   pure $ RemapPlayerResponse player.playerDisplayName
 
 adminModifyReportHandler :: ModifyReportRequest -> AppM NoContent
-adminModifyReportHandler ModifyReportRequest {rid, report} = case validateReport report of
+adminModifyReportHandler ModifyReportRequest {rid, timestamp, report} = case validateReport report of
   Failure errors -> throwError $ err422 {errBody = show errors}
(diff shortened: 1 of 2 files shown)


backend/src/AppServer.hs
```diff
@@ -238,14 +238,15 @@ adminRemapPlayerHandler RemapPlayerRequest {fromPid, toPid} = runDb $ do
...
+adminModifyReportHandler ModifyReportRequest {rid, timestamp, report} = case validateReport report of
   Failure errors -> throwError $ err422 {errBody = show errors}
   Success _ -> runDb $ do
     oldReport <- readOrError ("Cannot find report " <>: rid) $ lift . get $ rid
     Entity newWinnerId _ <- readOrError ("Cannot find player " <>: report.winner) $ getPlayerByName report.winner
     Entity newLoserId _ <- readOrError ("Cannot find player " <>: report.loser) $ getPlayerByName report.loser
+    let newTimestamp = fromMaybe oldReport.gameReportTimestamp timestamp
 
-    let newReport = toGameReport oldReport.gameReportTimestamp newWinnerId newLoserId oldReport.gameReportLogFile report
+    let newReport = toGameReport newTimestamp newWinnerId newLoserId oldReport.gameReportLogFile report
```
chelseaSchmidt, line 249: I don't know if I'm thinking about this right but since the new timestamp might shift a game before other existing games that used the player's Elo rating at that moment in time, should a timestamp change trigger a reprocess?

chelseaSchmidt (approved):
Easy!

backend/src/AppServer.hs
sirrus233: YES! Great catch.

chelseaSchmidt:
> Not sure how the UI should support this. If it's hard, we could punt there if need-be.

I'm thinking a combo of a date picker and time picker would make this not too much of a chore - then can translate that into one of the three valid string formats. Thanks for doing the research on that!

PR #141, opened 2025-01-08
Create database indices

### LINKED ISSUE #120: Indexes on database
Profile the reprocessing time

sirrus233:
Resolves #120 

(Note, this change will require a database migration in addition to a deployment of the server binary. Someday would be nice to be able to do this programmatically, but it's fine for now).

After profiling the times, I wouldn't call it revolutionary, but definitely worth doing. Measured runtime of a full reprocess dropped from ~13 seconds to ~ 9 seconds, which is about a 30% improvement.

I believe we can call this good, until/unless we need additional perf. So the following is just my musings on that problem:

-- 

Of the 9 seconds taken by a reprocess, they're split 4 / 5 between reading reports and processing them respectively. The 4 second time on reading is interesting, because it can't be reduced very easily with our current storage patterns, and is heavily CPU bound (the same operation takes ~1 second on my personal machine). The "next step" here would be the ability to only reprocess those games that happened AFTER the game where the edit occurred (this would make a reprocess entirely negligible, since most reprocessing is done on reports from the last month or two at most). To do this, we would have to revise how we store player stats - which would eventually amount to some kind of tradeoff between runtime and storage space. I think this would be a good thing to investigate - we would have to proceed carefully to make sure we don't impact fast display of the leaderboard.

### DIFF
--- a/backend/src/Types/Database.hs
+++ b/backend/src/Types/Database.hs
@@ -5,7 +5,7 @@ module Types.Database where
 
 import Control.Monad.Logger (LogLevel (..), ToLogStr (..))
 import Data.Time (UTCTime)
-import Database.Esqueleto.Experimental (Entity, runMigrationQuiet, runSqlPool)
+import Database.Esqueleto.Experimental (Entity, rawExecute, runMigrationQuiet, runSqlPool)
 import Database.Esqueleto.Experimental qualified as SQL
 import Database.Persist.TH (mkMigrate, mkPersist, persistLowerCase, share, sqlSettings)
 import Logging (Logger, log)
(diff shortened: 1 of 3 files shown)


backend/src/Types/Database.hs
```diff
@@ -80,6 +80,15 @@ migrateSchema dbPool logger = do
   unless (null migrations) (log logger LevelWarn "Database schema changed. Running migrations.")
   mapM_ (log logger LevelDebug . toLogStr) migrations
 
+  let index_statements =
```
sirrus233, line 83: Unfortunately, the Persistent DSL doesn't handle index creation, so we have to use raw/untyped SQL here.

chelseaSchmidt (approved):
4 seconds saved is HUGE, that can feel like a lifetime! Heck yeah

Good info to have on where the other 9 seconds goes, glad you did the profiling. I really like the idea of reprocessing only reports subsequent to the targeted game. Fun post-MVP mini project!

backend/src/Types/Database.hs
chelseaSchmidt: 🤮 (joking, this is nbd lol)

PR #143, opened 2025-01-08
Logfile Validation

### LINKED ISSUE #134: Validate file uploads are not scary

sirrus233:
Resolves #134 

Certainly not bulletproof, but I would argue good enough for now.

* On the frontent, accept text/plain or *.log extension files as input (I checked and confirmed it is Mac-specific nonsense going on with the logs. It knows they are text/plain, but browsers on Mac don't only get the MIME type. Mac has some notion of "hierarchy" that the browser is not aware of.)

* Block uploads > 1MB on the server

* Do a little bit of parsing validation on the backend, check at least for the header that should be present in all logs

### DIFF
--- a/frontend/src/FileUpload.tsx
+++ b/frontend/src/FileUpload.tsx
@@ -28,6 +28,7 @@ export default function FileUpload({
                     id={id}
                     style={{ display: "none" }}
                     type="file"
+                    accept=".log,text/plain"
                     onChange={(event) =>
                         onChange(event.target.files?.[0] || null)
                     }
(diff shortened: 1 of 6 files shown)


frontend/src/FileUpload.tsx
```diff
@@ -28,6 +28,7 @@ export default function FileUpload({
                     id={id}
                     style={{ display: "none" }}
                     type="file"
+                    accept=".log,text/plain"
```
chelseaSchmidt, line 31: Heyyy cool I didn't know you could do this! Easy peasy

chelseaSchmidt (approved):
Solid 👏 thanks for doing frontend too!

PR #146, opened 2025-01-08
Display Rated/Unrated instead of Ranked/Unranked

### LINKED ISSUE #145: Change Ranked/Unranked to Rated/Unrated

chelseaSchmidt:
Resolves #145 ~~(I don't know if you prefer to make this change at the data level, but this would be enough for MVP at least)~~


chelseaSchmidt:
Updated to make this change at the data level. I tested the frontend and backend locally, but didn't do a test migration - does that seem fine?

sirrus233 (approved):
Yes, looks great! Thank you for doing the backend. We'll do a data migration when we deploy this and that should take care of it

chelseaSchmidt:
Sweet 👌

PR #147, opened 2025-01-09
Periodically backup database

### LINKED ISSUE #108: Auto-backup of server data

sirrus233:
Resolves #108 

Backs up the database (compressed) to S3 every hour.

PR also contains a little bonus - a helper migration script to do spreadsheet migrations. Doesn't have anything to do with backup so PROBABLY should be in a separate PR. But it is here :shrug:


chelseaSchmidt (approved):
More bash land! Very cool 🎉

PR #152, opened 2025-01-20
Refactor getInitialFormData to be better

chelseaSchmidt:
This is the thing we paired on, have 0% tested it because I'm lazy and waiting for the prod server to be fixed instead of running the server locally, ~so don't merge 😬~ nvm this is what draft PRs are for


chelseaSchmidt:
Tested and ready for review now! Got tired of seeing it in the PR queue

sirrus233 (approved):
🚢

PR #157, opened 2025-01-26
Game reports frontend pagination

chelseaSchmidt:
I touched the backend, sorry to intrude, felt super out of my depth 😬 I needed the total record count and did it inefficiently by calling `getAllGameReports`, which throws away any backend time savings from pagination. Hoping you can help me fix it?

Anyway, the main frontend part is the `Pagination` component! Not too complicated but the logic for displaying page ranges split by the `...`s was fun.

I did local testing by mocking props and the total page count mostly; once the server code is synced up I'll definitely do real-life testing to make sure page numbers are for sure lining up with reports.

image

image

image

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -19,6 +19,7 @@ import Database
     getAllGameReports,
     getAllStats,
     getGameReports,
+    getNumGameReports,
     getPlayerByName,
     getPlayerStats,
     insertGameReport,
@@ -256,9 +257,9 @@ submitReportHandler (SubmitReportRequest rawReport logFileData) = do
 
 getReportsHandler :: Maybe Int64 -> Maybe Int64 -> AppM GetReportsResponse
 getReportsHandler limit offset = do
-    allReports <- runDb (getAllGameReports OldestToNewest)
+    total <- runDb getNumGameReports
     reports <- runDb (getGameReports limit' offset')
-    pure GetReportsResponse {reports = fromGameReport <$> reports, total = length allReports}
+    pure GetReportsResponse {reports = fromGameReport <$> reports, total}
     where
       maxLimit = 500
       (limit', offset') = case (limit, offset) of
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -53,7 +53,7 @@ import Servant.Server.Experimental.Auth (AuthServerData)
 import Types.Api
   ( DeleteReportRequest (..),
     GetLeaderboardResponse (GetLeaderboardResponse),
-    GetReportsResponse (GetReportsResponse),
+    GetReportsResponse (..),
     GoogleLoginResponse,
     IdToken,
     LeaderboardEntry (..),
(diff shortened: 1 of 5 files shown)


frontend/src/GameReports.tsx
```diff
@@ -49,6 +50,8 @@ const TABLE_TOP_POSITION =
     TABLE_BTN_HEIGHT_PX +
     TABLE_ELEMENTS_GAP * 2;
 
+const PAGE_LIMIT = 100;
```
chelseaSchmidt, line 53: Want to make this a dropdown menu in the future - will make a low priority issue for it

sirrus233 (approved):
Awesome stuff, love that you just took the initiative to make the component - I think it looks great! And now if we ever want to tweak it, that's easy too :D

Super happy :tada:

backend/src/AppServer.hs
```diff
@@ -203,13 +203,16 @@ submitReportHandler (SubmitReportRequest rawReport logFileData) = do
...
 
 getReportsHandler :: Maybe Int64 -> Maybe Int64 -> AppM GetReportsResponse
-getReportsHandler limit offset = GetReportsResponse . map fromGameReport <$> runDb (getGameReports limit' offset')
-  where
-    maxLimit = 500
-    (limit', offset') = case (limit, offset) of
-      (Nothing, _) -> (maxLimit, 0)
-      (Just lim, Nothing) -> (lim, 0)
-      (Just lim, Just off) -> (lim, off)
+getReportsHandler limit offset = do
```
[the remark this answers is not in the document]
chelseaSchmidt: Awesome!! I added this helper exactly as-is in `Database` (thank you for writing it), and practiced the fmap/applicative/bind stuff we discorded about in the implementation of `getReportsHandler`. The version I ended up committing was the nested bind, because I thought it looked the prettiest! Wouldn't have got there on my own, but next time an opportunity comes up will try either the bind or fmap/applicative approach myself as practice.

backend/src/AppServer.hs
```diff
@@ -255,13 +256,16 @@ submitReportHandler (SubmitReportRequest rawReport logFileData) = do
...
 getReportsHandler :: Maybe Int64 -> Maybe Int64 -> AppM GetReportsResponse
-getReportsHandler limit offset = GetReportsResponse . map fromGameReport <$> runDb (getGameReports limit' offset')
-  where
-    maxLimit = 500
-    (limit', offset') = case (limit, offset) of
-      (Nothing, _) -> (maxLimit, 0)
-      (Just lim, Nothing) -> (lim, 0)
-      (Just lim, Just off) -> (lim, off)
+getReportsHandler limit offset =
+    runDb getNumGameReports >>= \total ->
```
sirrus233, line 260: Very nice! Please do feel empowered to use do-syntax though, there are good reasons it exists! And one of those reasons is the awkwardness of requiring a bunch of nested binds when dealing with the results of multiple sequentially applied actions. 

What you had before wasn't wrong, and this isn't wrong either. Now that you see how this works, you are empowered to choose 😊

sirrus233 (approved):

backend/src/AppServer.hs
chelseaSchmidt: All the power 💪 💪

PR #160, opened 2025-01-27
Logout and UserInfo endpoints

sirrus233:
Finishes up the loose-ends from the initial auth PR

**Main additions**
* Logout API
* UserInfo API

`/logout` just nulls the sessionId in the auth database (recall that this is checked every time a protected action is attempted)

`/userInfo` returns a UserInfo response type... although this is quite boring at present, consisting of only a single `isAdmin` boolean - which should always be True if you get a successful response at all. We can use this to determine whether or not to show the admin controls on the page. The response can also be updated with more user info in the future, if that's ever needed.

**Other Refactor and Cleanup**
* Auth types moved to their own module, this solves some layering issues I had previously
* `UserId` and `SessionId` type aliases are now proper newtypes for better type safety
* Auth handler now passes along the authenticated user information to any protected handlers that need it (currently only `userInfoHandler` does)
* UUID values are handled more consistently

### DIFF
--- a/backend/src/Auth.hs
+++ b/backend/src/Auth.hs
@@ -3,42 +3,23 @@ module Auth where
 import AppConfig (AppM, Env, authCookieName, googleClientId, nt)
 import Crypto.JWT (JWTError (JWTExpired, JWTNotInAudience, JWTNotInIssuer))
 import Crypto.Random (getSystemDRG, withDRG)
-import Data.Aeson (FromJSON, eitherDecode)
+import Data.Aeson (eitherDecode)
 import Data.List (lookup)
 import Data.Text (isSuffixOf)
 import Data.Time.Clock.POSIX (getCurrentTime, posixSecondsToUTCTime)
 import Database (getAdminBySessionId, runAuthDb)
+import Database.Esqueleto.Experimental (Entity (..))
 import Jose.Jwa (JwsAlg (RS256))
 import Jose.Jwk (JwkSet (..))
 import Jose.Jwt (JwtContent (..), JwtEncoding (..), JwtError, decode)
 import Network.HTTP.Conduit (Manager, Response (..), httpLbs, parseRequest)
 import Network.Wai (Request (..))
-import Servant (AuthProtect, ServerError (errBody), err401, throwError)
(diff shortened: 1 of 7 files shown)


backend/src/Auth.hs
```diff
@@ -3,43 +3,24 @@ module Auth where
...
-import Servant (AuthProtect, ServerError (errBody), err401, throwError)
-import Servant.Server.Experimental.Auth (AuthHandler, AuthServerData, mkAuthHandler)
+import Servant (ServerError (errBody), err401, throwError)
+import Servant.Server.Experimental.Auth (AuthHandler, mkAuthHandler)
 import Types.Api (IdToken (..))
+import Types.Auth (Authenticated (..), GoogleIdTokenClaims (..), SessionId (..), UserId (..))
+import Types.Database (Admin (..))
 import Web.Cookie (parseCookiesText)
 
-data SessionIdCookie = SessionIdCookie -- Type-level tag for the auth scheme
```
sirrus233, line 21: This block is moved to the new Types.Auth module

backend/src/Types/Auth.hs
```diff
@@ -0,0 +1,29 @@
...
+
+import Data.Aeson (FromJSON)
+import Servant (AuthProtect)
+import Servant.Server.Experimental.Auth (AuthServerData)
+
+data SessionIdCookie = SessionIdCookie -- Type-level tag for the auth scheme
+
+type instance AuthServerData (AuthProtect SessionIdCookie) = Authenticated
+
+newtype UserId = UserId {unUserId :: Text}
```
sirrus233, line 11: Not sure if you're familiar with this idiom.

`newtype UserId = UserId {unUserId :: Text}` is the same as `newtype UserId = UserId Text` but with the added benefit of creating a function `unUserId :: UserId -> Text` that unwraps the newtype. The unwrapping function is named `un<MyType>` by convention.

This can be somewhat convenient in places where a pattern match unwrap might not be stylistically preferred - and you can see a usage in the Database module here.

chelseaSchmidt: Was not familiar with it, this is handy ✨

chelseaSchmidt (approved):
Straightforward, looks good!

PR #161, opened 2025-01-27
Fix EBS persistence

### LINKED ISSUE #155: Make EBS volume actually persistent

sirrus233:
Resolves #155

### DIFF
--- a/infrastructure/lib/infrastructure-stack.ts
+++ b/infrastructure/lib/infrastructure-stack.ts
@@ -104,15 +104,6 @@ export class InfrastructureStack extends cdk.Stack {
             machineImage: ami,
             vpc,
             role,
-            blockDevices: [
-                {
-                    deviceName: "/dev/sda2",
-                    volume: ec2.BlockDeviceVolume.ebs(8, {
-                        volumeType: ec2.EbsDeviceVolumeType.GP3,
-                        deleteOnTermination: false,
-                    }),
-                },
-            ],
             keyPair: ec2.KeyPair.fromKeyPairName(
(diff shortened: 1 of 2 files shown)


infrastructure/lib/infrastructure-stack.ts
```diff
@@ -104,24 +104,28 @@ export class InfrastructureStack extends cdk.Stack {
...
+            volumeType: ec2.EbsDeviceVolumeType.GP3,
+            removalPolicy: cdk.RemovalPolicy.RETAIN,
+        });
+
+        new ec2.CfnVolumeAttachment(this, "AttachServerStorage", {
+            volumeId: ebsVolume.volumeId,
+            instanceId: instance.instanceId,
+            device: "/dev/sda2",
+        });
+
```
chelseaSchmidt, line 126: What ended up being the difference between these two methods? Here is my blind approval though haha

chelseaSchmidt (approved):

infrastructure/lib/infrastructure-stack.ts
sirrus233: The difference between explicitly creating a new Volume object via `new ec2.Volume(...)` vs configuring one via parameters in the `Instance` object.

In the former case, a block storage device is created as a separate resource with CDK managing its life-cycle independently of the instance. In the latter, it has no existence of its own, so you get a new volume each time the instance is replaced.

chelseaSchmidt: Gotcha that makes sense!

PR #164, opened 2025-01-27
Fix buggy backup

### LINKED ISSUE #163: Investigate S3 database backups not working

sirrus233:
Resolves #163


chelseaSchmidt (approved):
I read the comment you left in the issue - for future reference should I think twice about deploying a non-prod branch, or would that only have applied for old branches that didn't include the backup infra change, so not an issue going forward?

sirrus233:
> should I think twice about deploying a non-prod branch

We should always think twice about this 🙃 

But in this case, the problem was likely the actually running of CDK synthesis on an old branch. I don't think we'd often do that.

Generally though, we want a better dev env. We have a post-MVP issue for that.

chelseaSchmidt:
Sorry, didn't mean to imply I'd be deploying dev branches willy nilly! Was thinking of those times when we deploy a branch sometimes during pairing. Makes sense and that clarifies!

PR #173, opened 2025-02-03
Display auth status in app header

chelseaSchmidt:
image

Expected behavior:
- If you're already signed in, the "signed in" message will be added to the header on mount if a request for your user info succeeds

- Otherwise, a successful sign-in with Google will kick off a new request for your user info, and the "signed in" message will be added to the header after the request succeeds

- If a user info request fails, nothing will be shown in the header

I couldn't do a full local E2E test, but I tested the UI by mocking out the Google button, and making my local server automatically validate auth requests without verification.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -6,6 +6,8 @@ import ExternalLink from "./ExternalLink";
 import GameReportForm from "./GameReportForm";
 import GameReports from "./GameReports";
 import Rankings from "./Rankings";
+import CheckIcon from "@mui/icons-material/Check";
+import CircularProgress from "@mui/joy/CircularProgress";
 import Box from "@mui/joy/Box";
 import IconButton from "@mui/joy/IconButton";
 import Drawer from "@mui/joy/Drawer";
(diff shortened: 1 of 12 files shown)


frontend/src/App.tsx
```diff
@@ -31,6 +36,16 @@ export default function App() {
         null
     );
 
+    const getUserInfo = () => {
+        setLoadingUserInfo(true);
+        axios
+            // .get("http://localhost:8081/userInfo")
+            .get("https://api.waroftheringcommunity.net:8080/userInfo")
+            .then((response) => setUserInfo(response.data))
+            .catch(console.error)
```
sirrus233, line 45: Might we consider catching/handling specific errors here? By inspecting the status code of the response?
Just a thought. See https://axios-http.com/docs/handling_errors

sirrus233 (approved):
Looks great :D

frontend/src/App.tsx
chelseaSchmidt: Absolutely, good call, done - and sorry for my scope creep 🤦‍♀️

chelseaSchmidt:
Added some new states to the header - these will only appear if the user deliberately tries to sign in using the Google button; they will not appear if the automatic on-mount auth status request fails.

**401 error:**

image

**Any error except a 401:**

image

frontend/src/App.tsx
chelseaSchmidt: I isolated the scope creep bit to https://github.com/sirrus233/wotr-community-website/pull/173/commits/31510a89cd525369f92fe1e1d91ed2104f30836e, so it might be easier to review this commit-by-commit now!

sirrus233:
Looks great, thanks for adding the extra error handling :)

🚢 🚀 ❗

PR #174, opened 2025-02-04
Hide admin controls unless user is authorized

chelseaSchmidt:


sirrus233 (approved):
Well that was easy! Awesome. This should close out the admin work!

Issue #176, opened 2025-02-04
Put Google login button into an admin menu


chelseaSchmidt:
In the admin menu, flip between login button and "signed in" message/"sign out" button

chelseaSchmidt:
Resolved by #247

PR #181, opened 2025-02-08
Put "admin" header above google sign-in

chelseaSchmidt:
Temporary stopgap to prevent confusion until we have more time for #176 

image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -339,6 +339,9 @@ function Home({
                 ))}
             </Section>
             <Section>
+                <Typography level="title-lg" mb={1}>
+                    Admin Sign-In
+                </Typography>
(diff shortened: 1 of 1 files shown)


sirrus233:
Possibly a suggestion for post-MVP, think about it:

The "You Cannot Pass" in the corner is cute - but we want to make sure we ~somehow~ defray the "danger" of this button, which is: we don't want non-admins thinking they have to use it. I wonder if we should show a reminder on a failed login that this is only for admins and normal users can report a game without a sign-in?

sirrus233 (approved):

chelseaSchmidt:
Yes! Was wondering if it might be confusing. I created a fast-follow milestone - I'll add an issue for it under that

PR #182, opened 2025-02-08
Tinker with landing page

chelseaSchmidt:
Sorry, this isn't MVP, I have no discipline! I liked the spirit of some of the landing page suggestions in the feedback doc and threw in these adjustments really quick. They aren't exactly as written in the doc, and not comprehensive - just added what I thought were a few small wins and leaving the rest for further discussion.

image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -9,9 +9,9 @@ import Rankings from "./Rankings";
 import CheckIcon from "@mui/icons-material/Check";
 import CircularProgress from "@mui/joy/CircularProgress";
 import Box from "@mui/joy/Box";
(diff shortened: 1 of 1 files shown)


sirrus233:
Overall, love this. A few thoughts:

* I don't know if I like the disclaimer being so prominent - a subtitle for the whole landing page is a pretty lofty position. Not sure where to move it though - maybe a footer?

* Consider making `Submit a game report` a different color? Or larger? I suggest this because that will be the most common reason for a user to visit the site right now.

sirrus233:
Admin sign-in could go in the footer too, frankly.

sirrus233 (approved):

PR #184, opened 2025-02-08
League points algorithm

sirrus233:
Algorithm specification doc: https://docs.google.com/document/u/0/d/1T2OIOfU9vlSJQd0xPoivFmHOUTvPOa1jjtaiN2vrHAQ/mobilebasic

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -40,7 +40,7 @@ import Database
 import Database.Esqueleto.Experimental (Entity (..), PersistStoreRead (..), PersistStoreWrite (..), Value (..), toSqlKey)
 import Logging ((<>:))
 import Network.HTTP.Client.Conduit (newManager)
-import Relude.Extra (groupBy)
+import Relude.Extra (groupBy, lookupDefault)
 import Servant
   ( AuthProtect,
     NoContent (..),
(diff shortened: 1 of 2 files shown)


backend/src/AppServer.hs
```diff
@@ -219,6 +219,25 @@ reprocessReports = do
...
+  where
+    baseScore = 0.1 * fromIntegral totalGames + fromIntegral totalWins
+    winRate = fromIntegral totalWins / fromIntegral totalGames :: Double
+    unplayedMultiplier
+      | (league == GeneralLeague || league == LoMELeague) && totalGames < 14 = 0
+      | (league == GeneralLeague || league == LoMELeague) && totalGames < 24 = 0.5 * winRate
+      | league == GeneralLeague || league == LoMELeague = winRate
+      | totalGames < 8 = 0
+      | totalGames < 18 = 0.5 * winRate
+      | otherwise = winRate
```
chelseaSchmidt, line 233: Good catch, I didn't notice in the spreadsheet (and didn't look closely at the doc before) that these thresholds differed by league. This matches up to what I see in the doc 💯

backend/src/AppServer.hs
```diff
@@ -219,6 +219,25 @@ reprocessReports = do
...
+    unplayedGames = fromIntegral . sum . map (\(wins, losses) -> 2 - wins - losses) $ stats
```
chelseaSchmidt, line 234: Halp! Got lost here, what's the logic of this calculation?

chelseaSchmidt (approved):

PR #193, opened 2025-02-12
Display country flags in leaderboard

chelseaSchmidt:


frontend/src/constants.ts
```diff
@@ -127,3 +127,41 @@ export const LEADERBOARD_START_YEAR = 2023;
 export const LEAGUE_START_YEAR = 2025;
 export const MAX_GAME_LOG_SIZE_MB = 1;
 export const MAX_GAME_LOG_SIZE_BYTES = MAX_GAME_LOG_SIZE_MB * 1024 * 1024;
+
+export const COUNTRY_FLAGS = {
```
sirrus233, line 131: Would it be worth it to pre-load the unicode code-points for all flags up front, so we don't need to update the frontend every time we get a new country?

https://apps.timwhitlock.info/emoji/tables/iso3166

sirrus233: And something like the download on this page might be useful, if we want to deploy our own images: https://flaglog.com/country-codes

sirrus233 (approved):

frontend/src/constants.ts
chelseaSchmidt: Much better! Wow, otherwise we'd have to change code every time an admin updated someone's flag, what was I thinking 🤦‍♀️ 

I'll start with the unicode in this PR, and add the second link to an issue for doing in a future PR 💯

PR #196, opened 2025-02-12
Add report links and higher resolution times to migration

sirrus233:
Links will be empty until the lift-n-shift script finishes, but they're correct!


backend/app/Migration.hs
```diff
@@ -32,16 +33,27 @@ import Validation (ReportError (..), validateReport)
 banList :: PlayerBanList
 banList = ["mordak", "mellowsedge"]
 
+insertLegacyEntry :: (MonadIO m, MonadLogger m) => ParsedLegacyLadderEntry -> DBAction m ()
```
sirrus233, line 36: I moved this from the Database module to avoid circular dependencies. It hasn't changed though.

chelseaSchmidt (approved):
🎉 🚢

PR #204, opened 2025-02-13
League delete remap

sirrus233:
What happened today:

A player in league named `Sepp G` wanted a name change to `SeppG`. Not knowing how to do that properly, an admin added a new player to the league named `SeppG`. We now have both `Sepp G` and `SeppG` in the system.

Ordinarily, this would be fixed with a `remap` operation, remapping `Sepp G` to `SeppG`. But remap wasn't written to work with league data. This changes things so it can.

We must consider the case where the "new" player, the result of the remap, already exists and is present in league (such as the above example). Although it is easy to imagine scenarios where that may not be the case.

As an added bonus, this PR also speeds up reprocessing after remap, if the "from" player didn't actually play any games.

### DIFF
--- a/backend/src/AppServer.hs
+++ b/backend/src/AppServer.hs
@@ -37,6 +37,7 @@ import Database
     runDb,
     updateActiveStatus,
     updateAdminSessionId,
+    updateLeaguePlayer,
     updatePlayerCountry,
     updatePlayerName,
     updateReports,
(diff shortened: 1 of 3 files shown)


backend/src/AppServer.hs
```diff
@@ -403,10 +404,11 @@ adminRemapPlayerHandler RemapPlayerRequest {fromPid, toPid} = runDb $ do
   _ <- readOrError ("Cannot find player " <>: fromPid) $ lift . get $ fromPid
   player <- readOrError ("Cannot find player " <>: toPid) $ lift . get $ toPid
 
-  updateReports fromPid toPid
+  updateCount <- updateReports fromPid toPid
+  updateLeaguePlayer fromPid toPid
   deletePlayer fromPid
   -- Warning: a deleted player that existed in pre-2022 data will be reintroduced via reprocessReports
-  reprocessReports
+  when (updateCount > 0) reprocessReports
```
chelseaSchmidt, line 411: Nice optimization 👏

chelseaSchmidt (approved):
💯 🚢

Issue #220, opened 2025-02-15
Folks seem to find the wording of Rated / Unrated to be terribly confusing


sirrus233:
I think people are confused because they think Rated means "tournament or league", and don't think of it as being a "normal" ladder game. So a lot of ladder games are getting mistakenly reported as Friendlies

sirrus233:
Not sure I'd consider this fully resolved yet. @bhilly's doc outlines a better solution

PR #221, opened 2025-02-15
Try to clarify rated game reporting

sirrus233:
Try to address #220 

Open to other thoughts here. See conversation: https://discord.com/channels/590789678890745857/818775910748258326/1340112563207405599

Having a perma-filled-in dummy value of "Ladder" next to the League/Tournament options might be a nicer way to illustrate this, rather than the additional text?

image


chelseaSchmidt (approved):
This is def less ambiguous - can iterate if it doesn't resolve confusion!

PR #224, opened 2025-02-15
Add migration script for posterity

sirrus233:
May we never need to run this again :)

(Feel free to rubberstamp, just seemed good to check it in)


chelseaSchmidt (approved):
Yes must for posterity! Rubber stamped ✅

PR #233, opened 2025-02-18
Remove special handling of Scotland and Wales flag images

chelseaSchmidt:
Fixed, thanks for making me think twice about it!


sirrus233 (approved):
Thanks :)

PR #235, opened 2025-02-20
Dev mode

chelseaSchmidt:
Not sure if what I've done is unusual, but the general internet consensus keeps pointing me to using `process.env` by injecting a simulated `process.env` into the browser, and I don't understand why common practice is to pretend a node thing exists in the browser. It seems misleading on what runtime environment we're expecting to be in. I opted for this instead:
- Have two possible `env` files (`env/prod.ts` or `env/dev.ts`)
- Define an `env` file path alias in webpack
- Point the alias to the proper file depending on the webpack build mode

### DIFF
--- a/backend/.vscode/settings.json
+++ b/backend/.vscode/settings.json
@@ -6,6 +6,7 @@
         "Angmar",
         "aragorn",
         "arwen",
+        "Baggins",
         "Barad",
         "bimap",
         "certbot",
--- a/backend/app/Server.hs
+++ b/backend/app/Server.hs
@@ -29,29 +29,19 @@ data AppMode = Dev | Prod
 corsMiddleware :: AppMode -> Middleware
 corsMiddleware mode = cors $ const $ Just policy
   where
-    policy = case mode of
-      Dev -> devPolicy
-      Prod -> prodPolicy
-
-    devPolicy =
+    origins = case mode of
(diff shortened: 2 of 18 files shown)


sirrus233:
I added some backend changes to this branch, under the guess that it would be good to push these things together (so they can be tested together as well).

Re: the env files - I see something similar is supported natively by react? Did you look at this documentation by any chance?

[StackOverflow](https://stackoverflow.com/questions/52896204/using-different-api-url-for-development-and-production-with-react-and-axios)
[React Docs](https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used)

chelseaSchmidt:
Sweet! I'll check out the backend stuff!

It looks like the native support is in Create React App, and from the stack overflow it looks like Create React App uses `dotenv` under the hood. It looks like I more or less recreated the functionality of the `dotenv` npm package, minus the simulation of `process.env` 🤷‍♀️

chelseaSchmidt:
Ran our changes all together locally and it works great, psyched that you added the backend. We could add bypassing of auth in a later PR? This is a vast improvement! Super ready to clear all my git stashes with various copies of commented out code 😂

chelseaSchmidt:
Auth added!

sirrus233 (approved):

PR #241, opened 2025-03-06
More relaxed logfile validation

sirrus233:
Should handle a few known cases where logfiles fail our validation:

* Windows style line endings
* Manually edited logs where passwords are inserted on the first lines
* Older client versions which can output some additional info prior to the "silent null" header we look for


chelseaSchmidt (commented):
💯 🎉

PR #244, opened 2025-03-15
Amazonka logging

### LINKED ISSUE #106: Replace Amazonka custom logger with standard app logger

sirrus233:
Resolves #106 

The Amazonka library (which handles all our AWS interactions for the backend) has its own logger, which doesn't interop with the logger we use for the rest of the app. This just provides a compatibility layer so that all the logging happens through our normal logger.

### DIFF
--- a/backend/src/Logging.hs
+++ b/backend/src/Logging.hs
@@ -59,8 +59,10 @@ fromAwsLogLevel AWS.Debug = LevelDebug
 fromAwsLogLevel AWS.Info = LevelInfo
 fromAwsLogLevel AWS.Error = LevelError
 
-toAwsLogger :: Logger -> AWS.Logger
-toAwsLogger logger awsLevel builder = log logger (fromAwsLogLevel awsLevel) (toLogStr builder)
+toAwsLogger :: LogLevelFilter -> Logger -> AWS.Logger
+toAwsLogger logFilter logger awsLevel builder = when (logFilter "" logLevel) (log logger logLevel (toLogStr builder))
+  where
+    logLevel = fromAwsLogLevel awsLevel
 
 log :: Logger -> LogLevel -> LogStr -> IO ()
 log logger = logger defaultLoc ""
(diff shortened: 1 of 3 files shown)


backend/src/Logging.hs
```diff
@@ -52,6 +53,17 @@ fileLogger path = do
...
   pure $ \_ _ level msg -> logger (\time -> formatLogMsg time level msg)
 
+fromAwsLogLevel :: AWS.LogLevel -> LogLevel
+fromAwsLogLevel AWS.Trace = LevelDebug
+fromAwsLogLevel AWS.Debug = LevelDebug
+fromAwsLogLevel AWS.Info = LevelInfo
+fromAwsLogLevel AWS.Error = LevelError
+
+toAwsLogger :: LogLevelFilter -> Logger -> AWS.Logger
+toAwsLogger logFilter logger awsLevel builder = when (logFilter "" logLevel) (log logger logLevel (toLogStr builder))
```
chelseaSchmidt, line 63: No productive feedback, just learning things! Was thrown because the function signature looks like it accepts 2 args, but there are 4 params?? And then looked at the types and realized AWS.Logger is a type of function that accepts 2 params, so when `toAwsLogger logFilter logger` is called in `Server.hs`, the result is a partially applied function matching the AWS.Logger function type. My Haskell lesson for the day ✔️

chelseaSchmidt (approved):
🚢

backend/src/Logging.hs
sirrus233: Yep! Basically just one of many ways to write a function that returns a function. I could have also written a two argument version that more explicitly defines and returns a different two argument function - perhaps that would have been clearer?

Nice work reading this and checking the type for more understanding when it looked weird. Your Haskell-fu has gotten so much stronger 💪

chelseaSchmidt: Ohhh it's like a decorator, that didn't even click when I wrote it out. And it was written totally clearly IMO, sometimes no amount of clarity can fix a newbie! Yay, thank you! 🙇‍♀️ 🎉

PR #245, opened 2025-03-15
Fix incorrect S3 key naming

### LINKED ISSUE #243: Game report links have the wrong name
Winner is always first and loser is always second - wrong args given to the function

sirrus233:
Resolves #243


chelseaSchmidt (approved):
💍 🌋

PR #246, opened 2025-03-16
Frontend leagues filter in game reports table

chelseaSchmidt:
| No selections, closed | Open | After clicking "Any league" | After clicking "No league" | With selections, closed |
|-|-|-|-|-|
| image | image | image | image | image |

- This was super quick to implement until I dove into the "any league" and "no league" options - hopefully establishing a pattern with this PR will make the other filters progressively faster!

- I didn't get fancy with the "any league" option; it's just a button that selects all the leagues for you (e.g. it doesn't deselect all leagues if you reclick, or flip to highlighted if you manually select all leagues - etc.)

- The "No league" option was weird. I included a backend adjustment for translating the meaning of an empty list, and also realized we don't have a way to filter for a mix of no-league and league - e.g. "game reports with no league OR with the WoME league". Maybe that would look like `[null, WoMELeague, ..etc]`, which would require more backend finagling than I wanted to do at the moment, and I don't think there's an urgent need 😬

### DIFF
--- a/backend/src/Database.hs
+++ b/backend/src/Database.hs
@@ -122,7 +122,9 @@ toFilterExpression report spec = foldr ((&&.) . fromMaybe (val True)) (val True)
             &&. ((report ^. GameReportWinnerId ==. val player2) ||. (report ^. GameReportLoserId ==. val player2))
     winnerFilter = toFilter report GameReportWinnerId spec.winners
     loserFilter = toFilter report GameReportLoserId spec.losers
-    leagueFilter = toFilter report GameReportLeague (map Just <$> spec.leagues)
+    leagueFilter = case spec.leagues of
+      Just [] -> Just (isNothing_ (report ^. GameReportLeague))
+      _ -> toFilter report GameReportLeague (map Just <$> spec.leagues)
 
 getAdminBySessionId :: (MonadIO m, MonadLogger m) => Text -> DBAction m (Maybe (Entity Admin))
 getAdminBySessionId sessionId = lift . getBy . UniqueAdminSessionId $ Just sessionId
--- a/frontend/src/TableFilter.tsx
+++ b/frontend/src/TableFilter.tsx
@@ -37,6 +37,30 @@ export default function TableFilter<O extends Option>({
         (o) => o !== undefined
     );
 
+    const isAllOption = (option: Option) =>
+        !!allOption && isOptionEqual(option, allOption);
+
+    const isEmptyOption = (option: Option) =>
+        !!emptyOption && isOptionEqual(option, emptyOption);
+
+    const isSpecialOption = (option: Option) =>
+        isAllOption(option) || isEmptyOption(option);
+
+    function translateSelections(selections: O[]): O[] {
+        const normalSelections = selections.filter((o) => !isSpecialOption(o));
+
+        if (selections.find(isAllOption)) {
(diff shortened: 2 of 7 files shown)


backend/src/Database.hs
```diff
@@ -122,7 +122,9 @@ toFilterExpression report spec = foldr ((&&.) . fromMaybe (val True)) (val True)
             &&. ((report ^. GameReportWinnerId ==. val player2) ||. (report ^. GameReportLoserId ==. val player2))
     winnerFilter = toFilter report GameReportWinnerId spec.winners
     loserFilter = toFilter report GameReportLoserId spec.losers
-    leagueFilter = toFilter report GameReportLeague (map Just <$> spec.leagues)
+    leagueFilter = case spec.leagues of
```
sirrus233, line 125: I worry slightly about the semantics of our FilterSpec type getting too tricky/complicated as we continue to add these kinds of corner cases - but I think this is excellent for now.

frontend/src/TableFilter.tsx
```diff
@@ -53,12 +63,37 @@ export default function TableFilter<O extends string | MenuOption>({
...
-                        onChange={(_, values) => onChange(values)}
+                        onChange={(_, values) => {
+                            const cleanedValues = values.filter((value) => {
+                                return !(
+                                    isOptionEqual(value, allOption) ||
+                                    isOptionEqual(value, emptyOption)
+                                );
+                            });
+
+                            onChange(
```
sirrus233, line 79: ⛏️ this might be easier to read as an extracted function with if/else instead of the chained ternaries.

frontend/src/TableFilter.tsx
```diff
@@ -83,3 +118,11 @@ export default function TableFilter<O extends string | MenuOption>({
         </th>
     );
 }
+
+function isOptionEqual(a?: Option, b?: Option) {
```
sirrus233, line 122: Why are the arguments to this equality checker optional? When might I want to compare something to nothing, or nothing to nothing?

sirrus233 (approved):
Some tough design decisions here, but overall I like them. I think this is a really great next step, and hopefully will generalize out to more filters. Awesome 🎉

backend/src/Database.hs
chelseaSchmidt: Sweet! Yeah totally see that, it doesn't feel super extensible. Will be on the lookout for refactor opportunities as we add more filters 💭

frontend/src/TableFilter.tsx
chelseaSchmidt: The `allOption` and `emptyOption` passed into `TableFilter` are optional; making the equality checker allow undefined options saves some repeated checking of whether or not `allOption` and `emptyOption` exist. (But actually in looking at it, only the second arg actually needs to be optional to get this benefit, so I'll remove the `?` from the first one!)

chelseaSchmidt: P.S. ended up making the second arg required too as part of adjusting the code for your other comment - it ended up a lot clearer I think 🎉

frontend/src/TableFilter.tsx
chelseaSchmidt: Fixed this - a little bit wordier now but a thousand times more readable 🫠

PR #253, opened 2025-03-19
Extract repeated HTTP request state management into a hook

chelseaSchmidt:
Some groundwork before #67 - extracts some repetitive state management code into a hook before sticking similar state management for the 4th/last network call into the app root.


sirrus233 (approved):
Nice!

PR #254, opened 2025-03-22
Handle params state inside useRequestState hook

chelseaSchmidt:
Little bit more cleanup before #67! Co-locates request params with rest of request state, and standardizes the way params are handled for each network call

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,7 +1,7 @@
 import axios from "axios";
 import { CssVarsProvider } from "@mui/joy/styles";
 import CssBaseline from "@mui/joy/CssBaseline";
-import React, { ReactNode, useEffect, useState } from "react";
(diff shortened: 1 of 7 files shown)


chelseaSchmidt:
Shucks, added one more commit 67ce819 to fix a sneaky bug I caused. Specifically, found out if I hit the "refresh" button on a page and the API server throws an error, causes a frontend runtime error 🤮

Root cause is there's a spot where an `onClick` prop passes an arg into a `refresh` function, and that arg fulfills the type of my optional `onError` handler in the `useRequestState` hook, so TypeScript didn't mind it. To safeguard against it being reintroduced, the fix forces the optional `onError` handler to be explicitly named when it's passed in

sirrus233 (approved):
LGTM! Nice catch on the error :)

Issue #130, opened 2025-01-07
Refresh leaderboard and game reports after any form submission


chelseaSchmidt:
Game reports are refreshed anyway for now, but after completing #67, will need to make it explicit

PR #255, opened 2025-03-22
Load game reports in app root

### LINKED ISSUE #67: Move `GET` endpoint requests to root app level to reduce network calls

chelseaSchmidt:
Resolves #67


sirrus233 (approved):
Looks good to me - but can you help with my remedial react.. This means that the first page of reports will be loaded upon hitting the app for the first time, and then again.. when?

chelseaSchmidt:
When a user:
- Hits the "refresh" button
- Changes the page
- Applies a filter
- In future, #130

I should check - is this the desired behavior?

PR #257, opened 2025-03-27
Add/update landing page links

### LINKED ISSUE #248: Add league rules to site
Either on homepage, league page, or both.

https://docs.google.com/document/d/1T2OIOfU9vlSJQd0xPoivFmHOUTvPOa1jjtaiN2vrHAQ/edit?usp=sharing

chelseaSchmidt:
Resolves #248 

Added `League Rules` under the `Rules Clarifications and FAQ` section, and also added a new `Technical Support` section with a couple links, since I've been meaning to and was in the area. And I updated the discord invite link - I think the current link is to an expired invite, but this new one doesn't expire.

I have plans to make the landing page prettier instead of a single column down the center like this - I like the current layout for mobile, but for desktop screen sizes, it would be cool to use a bit more of the horizontal space instead of needing to scroll.

image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -281,7 +281,7 @@ function Home() {
                     },
                     {
                         label: "Discord",
-                        href: "https://discord.gg/yZxQF4fK",
+                        href: "https://discord.gg/DUdcpdyNMm",
                         isDownload: false,
                     },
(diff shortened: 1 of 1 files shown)


sirrus233 (approved):
Thank you so much, Chelsea!

PR #259, opened 2025-03-29
Move leaderboard year selector adjacent to yearly stats header

### LINKED ISSUE #113: Make it more clear that year pertains only to annual stats

chelseaSchmidt:
Resolves #113 

I want the longer term solution to be (1) do my planned tables refactor, (2) make the filter bar currently used by game reports a reusable component for the leaderboard table, (3) make `year` a filter in the leaderboard filter bar. But this should do in the meantime.

image

image


sirrus233 (approved):
This is great. A big improvement with not too much effort. I agree it serves nicely for now.

PR #261, opened 2025-03-30
Display preselected "ladder" option for rated game type selection

### LINKED ISSUE #220: Folks seem to find the wording of Rated / Unrated to be terribly confusing

chelseaSchmidt:
Resolves #220 I think! Displays a preselected button that says "ladder" when "rated" is chosen. The button is permanently selected, and isn't actually hooked up to state; it's purely for UX.

image

### DIFF
--- a/frontend/src/GameReportForm/index.tsx
+++ b/frontend/src/GameReportForm/index.tsx
@@ -48,7 +48,7 @@ import Autocomplete from "../Autocomplete";
 import ErrorDisplay from "../ErrorDisplay";
 import FileUpload from "../FileUpload";
 import FormElement from "../FormElement";
-import MultiOptionInput from "../MultiOptionInput";
+import MultiOptionInput, { MultiOptionInputItem } from "../MultiOptionInput";
 import SelectNumericOptionInput from "../SelectNumericOptionInput";
 import SingleOptionInput from "../SingleOptionInput";
 import TextInput from "../TextInput";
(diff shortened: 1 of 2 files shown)


frontend/src/GameReportForm/index.tsx
```diff
@@ -272,6 +272,14 @@ function GameReportForm({
                         current={formData.competition.value}
                         onChange={handleInputChange("competition")}
                         validate={validateField("competition")}
+                        fakeOptions={
+                            <MultiOptionInputItem
+                                checked
+                                value="Ladder (required)"
```
sirrus233, line 278: 🚲 Do you think we need the "(required)"? Or is that sufficiently implied by the fact that you cannot untick the box? Could it be confusing?

sirrus233 (approved):

frontend/src/GameReportForm/index.tsx
chelseaSchmidt: On the fence, was thinking if someone tries to untick the box because they don't understand that "rated" == "ladder", and find they can't uncheck the box, they might think it's buggy as a result. Is that a realistic outcome? Or will people get it?

chelseaSchmidt: moved to discord

Issue #129, opened 2025-01-07
Leaderboard and Game Reports table styling


chelseaSchmidt:
- [x] Pin the column headers
- [ ] Cut off variable size cells with ellipses and display full contents in a hover tooltip or modal
- [x] Replace emojis with real icons?

chelseaSchmidt:
I was leaving this open to account for limiting the variable width of captured-settlement columns, but with the UI change in progress in #289, that column is going in a different direction UI-wise. So gonna go ahead and close 👍

Issue #128, opened 2025-01-07
Table filtering and searching


chelseaSchmidt:

chelseaSchmidt:
Suggestion: `e.g. all games that ira played in, all games between ira and bhilly, all T1 league matches from 2024, all games with an interest rating of 10, etc. and be able to mix and match those`

sirrus233:
Feature Request:
On the Game Results page, allow the visitor to filter/select different games, and then download just that. Or, have a few presets. For example, I'd love to be able to easily download:
Games with logs, AND
Games between players that are both rated 600 or higher as the side that they played, AND
Games with 2+ action tokens, AND
Games with FP military victories

etc.  Ideally, the filters would be:
Has a log
Minimum rating of at least one player (or both players), for the side they played in that game
Type of victory
Base vs. expansions
Turns taken
Interest rating
Date range of when the game was played
Player name

And then I can click "download" right there and I get all the logs in a zip.

sirrus233:
Further request from Ira, copied from discord:

On the game-reports page, allow the visitor to filter the reports. In approximate order of highest value filter to lowest value filter:
Game log exists (checkbox)
Game type (can select 1 or more)
Victory Type (can select 1 or more)
Current rating of winner (greater than, less than) XXXX   (or if it's easier, you could do rating of winner at time game was played)
Current rating of loser (greater than, less than) XXXX
Turn (greater than, less than, equal to) XX
Interest Rating (greater than, less than) XX
Initial eyes (greater than, less than, equal to) XX
Aragorn crowned (greater than, less than, equal to) XX
Tokens (greater than, less than, equal to) XX
Dwarven Rings (greater than, less than, equal to) XX
Player A name (matches with winner or loser)
Player B name (matches with winner or loser)
Timestamp (before, after, on)  (and if you're really fancy, then within +/- XX days of a specified date)

And then, the visitor can click a button to download all the logs that meet the filter. Or maybe the top X on this page, if there's a limit about downloading a nice zip file of logs.

ira212:
Seems like you might be making progress on this? Even if it's not fully done, it would be so great to get even a few more filters. For example, I recently played a game with 7 dwarven rings. I wanted to find other games with a high number of Dwarven Rings, but there's basically no way to do that.

Thank you for all your work on this! My main use case for the game records is to look at other games and try to learn from them. So being able to filter and search through them is super valuable to me. Thank you!

sirrus233:
[...]

PR #276, opened 2025-05-03
Use custom table component for game reports

### LINKED ISSUE #96: Fix tables rendering performance
Might be easiest to make a custom table component rather than using Material UI's

### LINKED ISSUE #168: Keep game report number and player columns on screen when scrolling right

chelseaSchmidt:
~Such draft,~ very refactor

~WIP, just opening as a draft so I can review easier!~

Resolves #168
Resolves #96
Makes progress on #129
Will make #128 ~a lot~ a little? easier

Sorry for this. Diff is a lot. I'm pleased with my table now. The only significant **visible** difference is pinned columns (pinned through `pairing` on desktop and through `edit` on mobile), which is a nice win; the rest is invisible refactoring.

Screenshots of the column pinning:

image

image

### DIFF
--- a/frontend/src/GameReports.tsx
+++ b/frontend/src/GameReports.tsx
@@ -1,17 +1,13 @@
-import React, { CSSProperties, ReactNode, useState } from "react";
+import React, { ReactNode, useState } from "react";
 import Box from "@mui/joy/Box";
-import CollapseIcon from "@mui/icons-material/KeyboardArrowDown";
 import EditIcon from "@mui/icons-material/EditTwoTone";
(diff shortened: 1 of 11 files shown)


chelseaSchmidt:
Converted from draft, ready for review but only as much as your interest extends, a lot of moving of code and CSS in this one 😬

sirrus233 (approved):
I finally looked this over, and it LGTM!!! It's obviously a lot of changes, just from line count, so I focused mainly on seeing if I noticed anything obviously wrong. If there's anything you specifically think I should zero-in on, let me know and I'll take a closer look. Otherwise... :ship: :rocket: :tada:

PR #287, opened 2025-07-25
Add reminders to README for when chelsea forgets stuff

chelseaSchmidt:


backend/README.md
```diff
@@ -1,8 +1,10 @@
-# Setup
+# Run
 
 - `cd` into this directory
 - `cabal update`
-- `cabal build` or `cabal run`
+- `cabal run server dev`
```
sirrus233, line 5: Does this actually work? I would think you'd need to add a `--` to signify the end of arguments to `cabal` and the start of arguments to the binary. So like:

`cabal run server -- dev`

chelseaSchmidt: It does work! I discovered that both work for me, idk why the `--` isn't needed

chelseaSchmidt: Let me just update it to be the proper one anyway though

sirrus233 (approved):

PR #288, opened 2025-07-27
More filters

sirrus233:
This is intended to start addressing #128 

I've enhanced the filter spec type to support filtering on most of the remaining GameReport fields. Unfortunately, I haven't tested it, mostly because there isn't a very easy way to do so without having a facility to send requests to myself with the new spec (better unit testing could cover this easily, but we're not setup to run unit tests against data from the database, because we're bad).

I sort of intended this PR to be a collab PR where the frontend could be build/tested along with the backend, and then deployed together.

Another note: currently this does not filter on strongholds, victory points, expansions, or competition type.

Strongholds, competition type, and expansions are all stored as serialized lists in our DB, rather than normalized into separate tables. So it's hard to write a good query. This was some laziness during the DB design phase,  which is biting now. Will need a future database migration to perform the normalization, and then we can do these queries.

Victory points are hard to query, because we actually compute them (from the strongholds) on the FE! Oops. so also probably out of scope for this PR.


backend/src/Database.hs
```diff
@@ -111,20 +116,83 @@ toFilterExpression ::
...
+      InequalityFilter EQ a -> (entity ^. field) ==. val a
+    toMaybeInequalityFilter ::
+      (PersistEntity e) =>
+      SqlExpr (Entity e) -> EntityField e (Maybe Int) -> Maybe InequalityFilter -> SqlExpr (Value Bool)
+    toMaybeInequalityFilter entity field = \case
+      Nothing -> isNothing_ (entity ^. field)
+      Just (InequalityFilter LT a) -> (entity ^. field) <. (just . val $ a)
+      Just (InequalityFilter GT a) -> (entity ^. field) >. (just . val $ a)
+      Just (InequalityFilter EQ a) -> (entity ^. field) ==. (just . val $ a)
+    toListFilter ::
```
sirrus233, line 134: This is what used to just be called `toFilter`

Note that the type of the function has changed - it no longer expects values wrapped in a functor, but rather just the values themselves. This whole function can then be fmapped inside whatever functor it needs to be, when it is called. That seemed cleaner to me.

backend/src/Database.hs
```diff
@@ -111,20 +116,83 @@ toFilterExpression ::
   SqlExpr (Value Bool)
 toFilterExpression report spec = foldr ((&&.) . fromMaybe (val True)) (val True) filterList
   where
-    toFilter entity field values = ((entity ^. field) `in_`) . valList <$> values
```
sirrus233, line 114: renamed, see below

backend/src/Database.hs
```diff
@@ -111,20 +116,83 @@ toFilterExpression ::
...
         (player1, Just player2) ->
           ((report ^. GameReportWinnerId ==. val player1) ||. (report ^. GameReportLoserId ==. val player1))
             &&. ((report ^. GameReportWinnerId ==. val player2) ||. (report ^. GameReportLoserId ==. val player2))
-    winnerFilter = toFilter report GameReportWinnerId spec.winners
-    loserFilter = toFilter report GameReportLoserId spec.losers
-    leagueFilter = case spec.leagues of
-      Just [] -> Just (isNothing_ (report ^. GameReportLeague))
-      _ -> toFilter report GameReportLeague (map Just <$> spec.leagues)
+    timestampFilter =
+      spec.timestamp <&> \case
```
sirrus233, line 166: If you don't recall `<&>`, it's just argument flipped `<$>`

backend/src/Database.hs
```diff
@@ -111,20 +116,83 @@ toFilterExpression ::
...
+      spec.leagues <&> \case
+        [] -> isNothing_ (report ^. GameReportLeague)
+        leagues -> toListFilter report GameReportLeague . map Just $ leagues
+    tokensFilter = toInequalityFilter report GameReportActionTokens <$> spec.tokens
+    dwarvenRings = toInequalityFilter report GameReportDwarvenRings <$> spec.dwarvenRings
+    corruption = toInequalityFilter report GameReportCorruption <$> spec.corruption
+    mordor = toMaybeInequalityFilter report GameReportMordor <$> spec.mordor
+    aragorn = toMaybeInequalityFilter report GameReportAragornTurn <$> spec.aragorn
+    treebeard =
+      spec.treebeard <&> \case
```
sirrus233, line 187: Sort of opinionated on this one: I'm assuming a filter of `Treebeard = False` should give all games in which Treebeard was available but not mustered, AND all games that didn't use the Treebeard expansion at all. But willing to change this if you prefer different semantics.

backend/src/Types/Api.hs
```diff
@@ -307,12 +307,42 @@ fromLeagueGameStatsMap playerId =
 
 type ExportResponse = (Headers '[Header "Content-Disposition" String]) (SourceIO StrictByteString)
 
+data TimestampFilter = Before UTCTime | After UTCTime | Between UTCTime UTCTime deriving (Generic)
+
+instance FromJSON TimestampFilter
+
+data InequalityFilter = InequalityFilter Ordering Int deriving (Generic)
```
sirrus233, line 314: From the base library, `data Ordering = GT | LT | EQ`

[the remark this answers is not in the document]
sirrus233: That's a good question. The `FromJSON` class is what defines a function to deserialize raw JSON data into a type that has an instance of the class. Our `FromJSON` instance for TimestampFilter is derived automatically for us - which is magic provided by the `deriving (Generic)` addendum. So we need to look up how Generic instances of FromJSON are defined on product types. I'll dig into that a bit

backend/src/Database.hs
chelseaSchmidt: Okay yes that makes sense! Not valuable in that case even if it were easy, I wasn't thinking realistically about the data

backend/src/Types/Api.hs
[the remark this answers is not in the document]
chelseaSchmidt: Switched to Discord but closing out here, thank you for the details! Will make sure to encode these as frontend types

PR #290, opened 2025-08-29
Support frontend inequality filtering and add turns filter

chelseaSchmidt:
First frontend filter to go with your backend work 🎉 

This PR establishes a new UI component for inequality filtering, uses that new filter for the `turns` column, and refines types to make them smarter at telling what properties are allowed/needed based on context.

This is branched from a rebased version of your [filters](https://github.com/sirrus233/wotr-community-website/tree/filters) branch. Should we just merge yours even though it hasn't been fully tested so we don't get into a feature branch situation? I did quick regression checks of the existing filters and everything seems A-OK so far.

| Closed | Operator open | Turn open | Filter applied
| - | - | - | - |
| image | image | image| image |

I went with a `Select` instead of an `Input` or `Autocomplete` for the turn input because I personally found scrolling and selecting to be an okay UX for our number columns with limited ceilings, but that might be a controversial choice 😬 I can adjust it if desired!

### DIFF
--- a/frontend/src/Table/InequalityFilter.tsx
+++ b/frontend/src/Table/InequalityFilter.tsx
@@ -1,0 +1,123 @@
+import React, { useEffect, useState } from "react";
+import Box from "@mui/joy/Box";
+import ClearIcon from "@mui/icons-material/CloseRounded";
+import IconButton from "@mui/joy/IconButton";
+import Option from "@mui/joy/Option";
+import Select from "@mui/joy/Select";
+import { InequalityFilter, InequalityOperator } from "../types";
+import { isDefined, range } from "../utils";
+import { TABLE_FILTER_HEIGHT } from "./constants";
+import { InequalityFilterProps } from "./types";
+
+export default function InequalityFilter({
+    current,
+    min,
+    max,
+    placeholder,
+    width,
+    onChange,
+}: InequalityFilterProps & { width: number }) {
+    const [currentOperator, currentValue = null] = current || [];
+
+    const [inputValue, setInputValue] = useState<number | null>(currentValue);
+    const [inputOperator, setInputOperator] = useState<InequalityOperator>(
+        currentOperator || "EQ"
+    );
+
+    useEffect(() => {
+        const didValueChange = currentValue !== inputValue;
+        const didOperatorChange = currentOperator !== inputOperator;
+
+        if (didValueChange) {
+            onChange(
+                isDefined(inputValue) ? [inputOperator, inputValue] : null
+            );
+        } else if (didOperatorChange && isDefined(inputValue)) {
+            onChange([inputOperator, inputValue]);
+        }
+    }, [currentOperator, currentValue, inputOperator, inputValue]);
+
+    const commonStyle = {
+        background: "white",
+        fontSize: "inherit",
+        height: TABLE_FILTER_HEIGHT,
+        maxHeight: "100%",
+        minWidth: 0,
+        minHeight: 0,
+        lineHeight: 0,
+    };
+
+    return (
+        <Box
+            display="flex"
+            alignItems="end"
+            justifyContent="center"
+            height="100%"
+            maxWidth={width}
+        >
+            <Select
+                size="sm"
+                value={inputOperator}
+                onChange={(_, o) => setInputOperator(o || "EQ")}
+                sx={{
+                    ...commonStyle,
+                    width: "fit-content",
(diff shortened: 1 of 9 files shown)


sirrus233:
> This is branched from a rebased version of your [filters](https://github.com/sirrus233/wotr-community-website/tree/filters) branch. Should we just merge yours even though it hasn't been fully tested so we don't get into a feature branch situation? I did quick regression checks of the existing filters and everything seems A-OK so far.

If this is rebased on top of the latest commit from that branch, do we need to merge that one? Or could we just merge this?

> I went with a Select instead of an Input or Autocomplete for the turn input because I personally found scrolling and selecting to be an okay UX for our number columns with limited ceilings, but that might be a controversial choice

In theory this is fine, but Turns doesn't actually have a ceiling. There's no theoretical maximum turn count for a game of War of the Ring, and in weird cases games can go very long. There is some upper bound that for all *practical* purposes nobody is likely to ever exceed, but it's certainly much higher than 13. It's hard to figure out our longest game historically, because the old form didn't track high enough either. The longest game I can find a reference to by searching discord is a 25 turn game played in 2022.

frontend/src/Table/types.tsx
```diff
@@ -1,14 +1,58 @@
...
+export type Option = string | MenuOption<unknown>;
+
+interface CommonFilterProps {
+    filterType: "autocomplete" | "inequality";
+    placeholder: string;
+    errorMessage?: ErrorMessage;
+    appliedCount: number;
+}
+
+export interface AutocompleteProps<O extends Option> extends CommonFilterProps {
```
sirrus233, line 14: ⛏️ Slight readability pref for `Opt` instead of just `O`?

frontend/src/Table/InequalityFilter.tsx
```diff
@@ -0,0 +1,123 @@
...
+            onMouseDown={(e) => e.stopPropagation()}
+            onClick={reset}
+            sx={{ minWidth: 0, minHeight: 0, lineHeight: 0, p: "1px" }}
+        >
+            <ClearIcon />
+        </IconButton>
+    );
+}
+
+function operatorLabel(operator: InequalityOperator) {
```
sirrus233, line 114: Possible enhancement: we could provide `>=` and `<=` inequalities on the FE, converting them into a filter expression. For example the `>= x` FE label converts to the `GT` and `EQ` operators under the hood.

sirrus233: Ah, or at least we could do that, if the API supported it, which it doesn't. Nevermind! If someone wants `>= x` they can just write `> x - 1`. Maths.

sirrus233 (approved):
Looks 🔥 🔥 🔥 
🚢

chelseaSchmidt:
> In theory this is fine, but Turns doesn't actually have a ceiling

Totally right, I made it go up to 50 to match how we handled this in the game report form, but no matter how you look at it it's messy to define an arbitrary ceiling. I'll update it to be a free number input!

frontend/src/Table/InequalityFilter.tsx
chelseaSchmidt: I like this and we can support it without changing the API! Can just have the FE do the `>= x` to `> x - 1` translation. Shouldn't be too much work 💯

sirrus233: Me, a backend engineer.

image

sirrus233:
I'm confused about the branch still. Why would we rebase my branch, if you have already rebased it to build on top of it? I may just be too tired to understand :)

In any case, I don't see any reason for all of this to remain unmerged. We should merge whatever we have that's ready, and work incrementally from there.

frontend/src/Table/InequalityFilter.tsx
chelseaSchmidt: 😂 💀

chelseaSchmidt:
Sorry haha, it's because the true #288 uses the old un-rebased branch. I just rebased a local copy and pushed that new branch to remote to act as my base. I didn't wanna touch your PR directly, entirely too rude 🙃

chelseaSchmidt:
TLDR - merge your PR and all will be solved, then I can rebase mine onto main

chelseaSchmidt:
Awesome, thank you, all rebased now, and gonna try to do my changes today so we can deploy 🏃‍♀️

PR #291, opened 2025-09-05
Add README notes on backend deployment

chelseaSchmidt:


backend/README.md
```diff
@@ -7,15 +7,48 @@
...
+  - [Install the Session Manager plugin for AWS CLI](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html)
+  - Add the following to `~/.ssh/config`:
+    ```
+    Host api.waroftheringcommunity.net
+      User ec2-user
+      IdentityFile ~/.ssh/serverkey.pem
+      HostName <instance_id>
+      ProxyCommand sh -c "aws ssm start-session --profile wotrcommunity --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
+    ```
+    Where `<instance_id>` can be found in the EC2 console, but don't know about `IdentityFile` origin, forgot :'D
```
sirrus233, line 32: -- To answer your question --

The IdentityFile contains a private key matched to the public key stored on the EC2 instance itself. These keys were generated as a matched pair (by me, manually), and cannot be regenerated. So we have the only two copies, and would need to (as securely as "possible") transfer the key file to anyone else if it became necessary for them to have SSH access.

-- For educational purposes --

This is obviously an unfortunate attack surface, since passing around key files opens you up to the risk that they may be intercepted. And of course if the key is compromised or needs to be rotated, *everyone* will lose their access all at once. In general this method of handling credentials is considered unwise and inconvenient. 

The better mechanism would be - instead of a single shared private key - every client who needs to access the server should generate their own public-private key pair, and then give their unique public key to the server (one-time upload). That key gets added to an internal keyring on the server, so it can recognize connection attempts from the matched key in the future. You've probably actually done this before, say, to connect to github over ssh.

Transferring the *public* key over the internet is safe, because that's not intended or required to be kept secret - only the private key is. And now if your private key gets compromised somehow, the blast radius is limited to you and you alone.

All that said, our "less good" approach is mostly fine for only two users, and doesn't even carry as much security risk as it otherwise would, because we don't have an SSH service exposed on any externally accessible port. The only way to make an SSH connection to our server is to pass through AWS SSM, which means that even if our key became compromised, an attacker would still have to compromise our AWS IAM roles to actually access the server. We COULD even drop SSH authentication entirely, in which case the PEM file is not needed, but as we already discussed, it's somewhat convenient to be able to use standard tools that utilize ssh tunneling (e.g. scp)

So, you know, don't leak both the shared PEM file *and* your AWS secret key please :)

sirrus233 (approved):
Thank you for updating docs!!! If you ever want to do this in the future, please feel free to push without code review :)

PR #292, opened 2025-09-07
Add batch of frontend inequality filters

chelseaSchmidt:
Should be all the non-maybe inequality filters! I'm thinking about following up with changing text alignment to be left-aligned instead of centered, because right now the columns are wider than really needed to avoid overlap between filter symbols and header text. But I didn't want to mix the new filters with UI tinkering, so for now, the columns are just a bit wide.

## Initial

image

image

## With filters applied

image

image


frontend/src/GameReports.tsx
```diff
@@ -201,18 +201,82 @@ export default function GameReports({
...
+            },
+        },
+        {
+            key: "Corruption",
+            width: 150,
+            filter: {
+                filterType: "inequality",
+                placeholder: "Corruption",
+                min: 0,
+                max: 18,
```
sirrus233, line 238: Although corruption certainly has *some* theoretical upper bound, I'm not entirely sure what it is, and it's definitely higher than 18.

I could imagine a (strange) world where Fellowship moves on the Mordor track while at 11 corruption with Gollum as guide. Shadow has all 10 of their action dice allocated as Eyes in the hunt box. This is the Fellowship's 6th move this turn, having rolled C's on all their dice (they have GtW and Aragorn), but they haven't dunked yet because they hit some red tiles along the way. An eye is drawn.

In this scenario, the Fellowship would take 15 corruption damage from the move attempt, ending the game in a SPRV with Fellowhip at 26 corruption total.

That might be the theoretical limit in the Base game (??) Although in LOME, it could be even higher. Probably not by much. Without thinking about it too, too much I doubt corruption can probably exceed ~30?

chelseaSchmidt: Good catch, I pulled the limit from the game report form and didn't do the math! Will update to 30 and also make the game report form use the same ceiling.

P.S. It would be really exciting to see this happen in a game 😆

chelseaSchmidt: Confirmed, took a look and the only BE validation on corruption I'm finding is that it's 0 or higher, or if it's a shadow ring victory, that it's 12 or higher

chelseaSchmidt:
I can't decide if the latest changes are big enough to be worth re-review - they're not super interesting, but will leave this open for you to do a last sign off before I merge! Then maybe a release update in discord is due ✨

PR #293, opened 2025-09-07
Display definition of active/inactive player in rankings page

### LINKED ISSUE #201: definition of active v inactive?
With that being a more prominent choice on the rankings, I'm realizing I don't actually know what "active" means.... is there some kind of tooltip definition that could be added, or a dictionary elsewhere?

chelseaSchmidt:
Resolves #201

| Initial | On hover |
| - | - |
| image | image |

### DIFF
--- a/frontend/src/Rankings.tsx
+++ b/frontend/src/Rankings.tsx
@@ -128,16 +128,25 @@ function Rankings({
                             enterTouchDelay={0}
                             size="sm"
                             title={
-                                <Box>
-                                    <Box mb="5px">
+                                <Box
+                                    display="flex"
+                                    flexDirection="column"
+                                    gap="5px"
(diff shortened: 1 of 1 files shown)


sirrus233 (approved):
Omg, thank you for getting to this!! Emily actually filed this one 😆 

Really just one minor nit, if it's easy to do. Is it possible to render a ligature for `>=` instead of the two distinct characters? How does that work?

chelseaSchmidt:
> Is it possible to render a ligature for >= instead of the two distinct characters?

Wow. First, TIL what the word ligature means in this context, thank you. Second, I straight up forgot that the normal-person symbol is ≥. I'm a space cadet 🤦‍♀️ The answer is yes, there's a text character for it and everything, so just a copy change!

PR #298, opened 2025-09-12
Support frontend boolean filtering and add game log filter

chelseaSchmidt:
Game log column filter, underpinned by a `BooleanFilter` component which can be reused for the comments and treebeard columns.

| Initial | Menu open | "No log" selected | "Log" selected |
| - | - | - | - |
| image | image | image | image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -120,6 +120,7 @@ export default function App() {
                 corruption: null,
                 initialEyes: null,
                 interestRating: null,
+                hasLog: null,
             },
         },
         sendRequest: (params) => {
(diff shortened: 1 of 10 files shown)


sirrus233 (approved):
LGTM! Feel free to toss the Treebeard filter in there as well, since that's the other boolean one.

FYI (for Treebeard): Our database makes a distinction between Treebeard `true` (he was mustered`, `false` (he was played with, but not mustered), and `null` (the treebeard expansion wasn't being used). But the *filter* logic does not distinguish, instead interpreting a filter value of `false` to mean `false | null` on the backend. 

I think this is probably fine, especially once we allow filtering by expansions used. Especially considering the fact that Treebeard is almost never mustered, and the most relevant filter folks may be interested in is `Treebeard = true`

PR #301, opened 2025-09-14
Frontend victory type filter

chelseaSchmidt:
Victory type filter! There are two commits to separate out some type cleanup from the new filter. There's a type in one of the autocomplete filter params that's not being inferred properly, which affected all the prior autocomplete filters too, but hadn't got around to fixing it til now.

| Initial | Menu open | Selections in progress | Filters applied |
| - | - | - | - |
| image | image | image | image |

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -114,6 +114,7 @@ export default function App() {
                 winners: [],
                 losers: [],
                 leagues: [],
+                victory: [],
                 turns: null,
                 tokens: null,
                 dwarvenRings: null,
--- a/frontend/src/GameReports.tsx
+++ b/frontend/src/GameReports.tsx
@@ -10,7 +10,13 @@ import ModalDialog from "@mui/joy/ModalDialog";
 import Tooltip from "@mui/joy/Tooltip";
 import freeIconPath from "./assets/ring-emoji.png";
 import shadowIconPath from "./assets/volcano-emoji.png";
-import { ErrorMessage, GAME_LIMITS, leagues } from "./constants";
+import {
+    ErrorMessage,
+    GAME_LIMITS,
+    leagues,
+    sides,
+    victoryTypes,
+} from "./constants";
 import useMediaQuery from "./hooks/useMediaQuery";
 import { RefreshRequest } from "./hooks/useRequestState";
 import {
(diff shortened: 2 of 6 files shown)


sirrus233 (approved):
Overall this looks good (I'd be willing to see it merged) - but thinking about it at the app level, I wonder if we want to expose some more powerful features as part of this filter?

There are a few kind of powerful options that are hard to do with this filter as-is. What if I want to see just all Free victories (of any type). Or all Ring/Corruption victories (for any side)? What about all Concessions?

I think we could support this actually relatively easily, with a small Backend change (which could be included in this PR). Instead of the filter type being a tuple of `(side, victory)`, we'd want a new custom data type. I'm thinking something like:

```
data VictoryFilter = 
  VictorySideFilter Side |
  VictoryKindFilter Victory |
  VictoryFilter Side Victory
 ``` 
 
Then we just give it a FromJSON instance for deserialization, and fix up the lambda for the fmap in the database filtering code. This:

```
    victoryFilter =
      spec.victory
        <&> foldr1 (||.)
        . fmap (\(s, v) -> (report ^. GameReportSide) ==. val s &&. (report ^. GameReportVictory) ==. val v)
        ```
        
Would just need to be modified such that the lambda is now a case statement that disambiguates on the value of the VictoryFilter type.

What do you think? Do you like this idea? Not like it?

sirrus233:
IF you do decide you wanna take a crack at my backend suggestion, [this](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/lambda_case.html) is a super nice bit of syntactic sugar that could be used.

backend/src/Database.hs
```diff
@@ -171,9 +170,17 @@ toFilterExpression report spec = foldr ((&&.) . fromMaybe (val True)) (val True)
...
+        foldr
+          ( (||.)
+              . ( \case
+                    VictoryKindFilter k -> report ^. GameReportVictory ==. val k
+                    VictorySideFilter s -> report ^. GameReportSide ==. val s
+                    VictoryComboFilter s k -> (report ^. GameReportSide) ==. val s &&. (report ^. GameReportVictory) ==. val k
+                )
+          )
+          (val False)
+          (v : vs)
```
chelseaSchmidt: This works technically but looking at the diff I think I can make it cleaner, will work on it

sirrus233:
👍

PR #303, opened 2025-10-22
Support null in inequality filtering and filter on mordor/aragorn

chelseaSchmidt:
I'm not sure if I took the backend in the right direction or not in this PR. I couldn't figure out how the API could distinguish between `Nothing` and `Just Nothing` in JSON - I think they both serialize to `null`. So I made a new data type `NullFilter` that could be used in place of `Just Nothing` when we want to filter on the presence of `null`. Hopefully I didn't way overcomplicate it.

This is what the frontend would look like:

|  |  |  | 
| - | - | - |
| image | image | image |

### DIFF
--- a/backend/src/Types/Api.hs
+++ b/backend/src/Types/Api.hs
@@ -317,7 +317,7 @@ instance FromJSON InequalityFilter
 
 data NullableFilter f = NullFilter | ValueFilter f deriving (Generic)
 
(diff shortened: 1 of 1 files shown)


sirrus233 (approved):

chelseaSchmidt:
P.S. I have to run soon but should be able to kick off a deployment sometime this evening, FYI

PR #306, opened 2025-10-23
Relabel "Ring" victory kind filter option to "Ring/Corruption"

chelseaSchmidt:
Looking at this with fresh eyes after deploying and the "Ring" option feels potentially confusing. The option currently labeled "Ring" in the victory type filter is meant to encompass any ring-type victory, but might be confused to only represent a free ring victory.

| Old | New |
| - | - |
| image | image |


sirrus233 (approved):
Love it 👍

PR #310, opened 2025-10-31
GameReports module

chelseaSchmidt:
~(Draft for now)~

I let the size of `GameReports` get outta control 🥲 This change makes it its own module, splits the code into a few different files, and brings some game-report-specific components inside the new module.

At some point need to do the same thing with `Rankings` and `Leagues`, and then move some single-use form components into the `GameReportForm` module.

### DIFF
--- a/frontend/src/components/GameReports/rows.tsx
+++ b/frontend/src/components/GameReports/rows.tsx
@@ -5,6 +5,8 @@ import EditIcon from "@mui/icons-material/EditTwoTone";
 import IconButton from "@mui/joy/IconButton";
 import Tooltip from "@mui/joy/Tooltip";
 import ViewIcon from "@mui/icons-material/Visibility";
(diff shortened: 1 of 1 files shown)


frontend/src/components/GameReports/rows.tsx
```diff
@@ -0,0 +1,287 @@
...
+        })
+    );
+}
+
+interface FixedWidthTextProps {
+    children: ReactNode;
+    width: number;
+}
+
+function FixedWidthText({ width, children }: FixedWidthTextProps) {
```
chelseaSchmidt: I want to split the 3 sub-components here into their own files too, but trying to be marginally considerate about the size of the diff! Unless you object to more component files, I'll tack on a commit to split them off after you review/before I merge it (don't feel like that part needs a second set of eyes).

PR #314, opened 2025-11-07
Make app minimally usable on mobile

chelseaSchmidt:
~(Draft for now, need to double check stuff with fresh eyes later)~

This started as cleanup, but I ended up going down a rabbit hole and resolving #95:

- Mainly the game report form:
  - Make the Mordor steps help icon a sibling of the field label, instead of a child, so that touching it on mobile doesn't open the tooltip and the menu simultaneously. It's alignment also looks better this way, when the text starts to wrap

  - Below a certain breakpoint, vertically align the file upload element's children

  - Below certain breakpoints, reduce the form's margin and padding

  - Explicitly define the size of the image in the Mordor steps "help" tooltip so that the tooltip positions itself correctly based on available screen space

  - Below a certain breakpoint, vertically align radio buttons

- Add a minimum width to the app body of `285px`, so that we don't have to think about anything being squeezed further than that, and just introduce a scrollbar instead

- Below a certain breakpoint, vertically align the league selectors
  - To do this cleanly, also gave tables a dynamic height computation (which in turn required adding `height: 100%` to various levels of the DOM tree), as the previous manual height calculation wasn't flexible enough for this

### DIFF
--- a/frontend/src/components/ButtonSelector.tsx
+++ b/frontend/src/components/ButtonSelector.tsx
@@ -34,7 +34,7 @@ export default function ButtonSelector<T extends string | number>({
                 ...style,
             }}
         >
-            <ButtonGroup style={{ height: `${sizes.buttonSelectorHeight}px` }}>
+            <ButtonGroup sx={{ height: "36px" }}>
                 {options.map((option) => (
                     <Button
                         key={option}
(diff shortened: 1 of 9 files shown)


chelseaSchmidt:
@sirrus233 This expanded beyond housekeeping, but it's pretty CSS-heavy rather than introducing new code/logic. Do you want an opportunity to review or should I go ahead and merge?

PR #334, opened 2026-01-04
Order game report form strongholds

### LINKED ISSUE #323: Rearrange the order of the cities and strongholds that appear in the game report
Image
This is really trivial, but what the heck order are those in?! I can't figure it out.

I think East to West, North to South might make sense? When I'm thinking about the play of the game, I have a sense of which part of the board had the most action.

There are many reasonable orders. Here's an order that would be more pleasant to me personally:
Woodland Realm, Dale, Erebor  (these three often go together, and they are likely to see action)
Lorien
Helm's Deep, Edoras
Minas Tirith, Pelargir, Dol Amroth
Rivendell
The Shire
Grey Havens

I think that might also sort the more likely locations to the left. It's fine to have them all in a row; I was just grouping them like that because of how it made sense to me.

Alternately, tell me the logic of the current sort order, and I will be at peace. :) Thanks for all the work you do!

chelseaSchmidt:
Resolves #323 and #302 (unless we want to keep an issue open for future color coding? In which case I'll leave #302 open)

I decided against sharing the sort order with badges in historical game reports, because the badge layout has a really specific rhyme and reason that doesn't need to lock in the order of the game form's buttons with it.

I also decided against just reordering the root constant, because then it would be kind of secret/undocumented that we're relying on that list to stay ordered. So this is what I landed on.

### DIFF
--- a/frontend/src/components/GameReportForm/index.tsx
+++ b/frontend/src/components/GameReportForm/index.tsx
@@ -32,6 +32,7 @@ import {
     GameFormData,
     GameReportPayload,
     ProcessedGameReport,
+    SameElements,
     ServerErrorBody,
     ServerValidationError,
     Stronghold,
@@ -745,8 +746,8 @@ function validationErrorToMessage(
     }
 }
 
-function orderStrongholds(strongholds: Stronghold[]) {
-    const gameFormOrder: Stronghold[] = [
+function orderStrongholds(shs: Stronghold[]) {
+    const gameFormOrder = [
         "MinasTirith",
         "Pelargir",
         "DolAmroth",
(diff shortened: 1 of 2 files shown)


frontend/src/components/GameReportForm/index.tsx
```diff
@@ -744,3 +744,39 @@ function validationErrorToMessage(
             return "invalid stronghold selections for the indicated expansions";
     }
 }
+
+function orderStrongholds(strongholds: Stronghold[]) {
+    const gameFormOrder: Stronghold[] = [
+        "MinasTirith",
```
chelseaSchmidt: Just went back to the ticket and remembered Ira had a specific order in mind - I can update this to match the suggested order, oops. Unless you think keeping consistent with the order used by historical game reports is of higher priority

sirrus233 (approved):
I think the new order is probably fine, but maybe @ira212 can weigh in before we merge?

PR #344, opened 2026-02-15
Make prettier a local dependency and add helper scripts

chelseaSchmidt:
Wondering if the source of the formatting issues was that we don't have Prettier in the project as a dependency, so it was relying on my local VSC extension to get the version, instead of a fixed version. Almost all the issues look like `trailingComma` and ternary indentation, which I saw warnings about Prettier changing the defaults for at some point, but don't remember when. Either that, or I had a user-level override in my settings that's gone now 🤷‍♀️ 

Anyway, this adds Prettier as a dependency, a script for local auto-fixing, and a script that CI can use, when I set that up next.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -26,6 +26,7 @@
                 "css-loader": "^7.1.2",
                 "html-webpack-plugin": "^5.6.0",
                 "mini-css-extract-plugin": "^2.9.0",
+                "prettier": "3.8.1",
                 "ts-loader": "^9.5.1",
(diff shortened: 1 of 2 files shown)


sirrus233 (approved):

chelseaSchmidt:
@sirrus233 Sick! I'm gonna close this PR in favor of reviewing yours. I would still like to have formatting in CI, on top of a pre-commit hook. But the package.json script for that can be added in the same PR as CI, since your pre-commit hook will provide a much better local tool in the meantime. Thank you 🎉

PR #354, opened 2026-03-22
Formatting cleanup

sirrus233:
Formatting PR - this is the result of running `pre-commit run -a`, which runs our pre-commit checks against every file in our repo.

Submitting for review so you can check that things look how you want/expect. If anything looks wacky, we can adjust the new `.prettierrc.yaml` file at the repo root.


.prettierrc.yaml
```diff
@@ -1,3 +1,3 @@
-trailingComma: 'es5'
+trailingComma: "es5"
 tabWidth: 2
```
chelseaSchmidt, line 2: I'm noticing from this PR's indentation changes that this setting will put us in a halfway state where some files will have indentation of 2 and some 4. Should we set tab width to 4 here, and then in a separate PR, make simultaneous updates to make tab width 2 in `.prettierrc.yaml`, `frontend/.prettier.rc.js`, and `.vscode/settings.json` (maybe add a new line to `.zed/settings.json` too)?

I do prefer tab width of 2 in general, but wasn't looking forward to the massive PR 😬

chelseaSchmidt (commented):
Nice, I did my pre-commit setup from the README! Looking forward to having everything be enforced

